PR #34, opened 2026-05-27
smoke: 2-commit no-op against critical-issues-only — exercises matrix detect + pipefail-fix

thrillmot:
Smoke test for notify-clud-bug.yml's on:push trigger (PR #29) and the pipefail zero-grep-match fix (PR #32). Two commits both touching `skills/critical-issues-only/SKILL.md`: add a marker, then revert it. Net file content = identical to main. Once merged via squash, the squash-commit will touch the file path (path filter fires) BUT the diff has zero net change → notify-clud-bug.yml's detect job grep finds zero baseline matches → with A.8's `|| true` guard, gracefully emits empty matrix → no clud-bug PR opens. If anything ELSE happens, A.8 or matrix detect has a bug.

### DIFF
--- a/skills/critical-issues-only/SKILL.md
+++ b/skills/critical-issues-only/SKILL.md
@@ -1,4 +1,3 @@
-<!-- smoke test marker -->---
 name: critical-issues-only
 description: PR review discipline - flag only correctness, security, and performance issues. Skip nits.
 review_mode: shared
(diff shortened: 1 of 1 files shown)


thrillmot:
Closing — the sed I used to create a two-commit no-op accidentally stripped the YAML frontmatter opener (`---`), so the smoke isn't actually a no-op. Will retry with a cleaner approach (probably a trailing-newline touch via printf, not sed).

PR #35, opened 2026-05-27
smoke: 2-commit no-op via git revert — exercises matrix detect + pipefail-fix

thrillmot:
Smoke test for notify-clud-bug.yml's on:push trigger (PR #29) and the pipefail zero-grep-match fix (PR #32).

Two commits:
1. Add a marker line to `skills/critical-issues-only/SKILL.md`
2. `git revert HEAD --no-edit` to undo

Net file content = byte-identical to main. Once squash-merged, the squash commit touches the file path (path filter fires) BUT the diff has zero net change in baseline content → notify-clud-bug.yml's `detect` grep finds zero baseline matches → A.8's `|| true` guard kicks in → graceful empty-matrix no-op → no clud-bug PR opens.

Expected: workflow runs to completion, posts a notice about zero baselines, no PR appears on clud-bug.

If the workflow fails red, A.8's pipefail fix has a bug. If it opens a PR on clud-bug, the detect-step diff logic is wrong.

### DIFF
--- a/skills/critical-issues-only/SKILL.md
+++ b/skills/critical-issues-only/SKILL.md
@@ -1,3 +1,4 @@
+<!-- smoke marker -->
 ---
 name: critical-issues-only
 description: PR review discipline - flag only correctness, security, and performance issues. Skip nits.
--- a/skills/critical-issues-only/SKILL.md
+++ b/skills/critical-issues-only/SKILL.md
@@ -1,4 +1,3 @@
-<!-- smoke marker -->
 ---
 name: critical-issues-only
 description: PR review discipline - flag only correctness, security, and performance issues. Skip nits.


thrillmot:
Closing — realized this approach can't actually smoke-test the matrix detect path. The org ruleset enforces squash-only merges. A squash of two commits with net-zero diff collapses to a single commit with no path-touching diff → notify-clud-bug.yml's `on: push: paths: [skills/.../SKILL.md]` filter doesn't match → workflow doesn't fire at all. The smoke would silently no-op for the wrong reason.

Real options to exercise A.8's pipefail-safe path:
1. Admin-bypass push 2 commits directly to main (a force-push-rewrite scenario; ugly)
2. Make a real change to a baseline → workflow fires → opens a clud-bug PR (tests matrix+sync, NOT the zero-match path A.8 fixes)
3. Defer until a real baseline change happens organically; A.8's fix is logical-correct and will exercise on the first push-event that satisfies the path filter but produces zero git-diff matches (e.g. force-push rewrite)

Going with option 3.

PR #51, opened 2026-05-29
feat: add applies_to to 3 scoped skills (companion to clud-bug 0.0.K / v0.6.21)

thrillmot:
## Summary

Companion to thrillmade/clud-bug#113 (0.0.K). Adds `applies_to:` blocks to three dedicated skills so they only load on PRs whose changed files match their scope. Leaves `pii-and-compliance` un-scoped (universal) — PII leaks can hide in any code path that emits data outward, and the false-negative cost of skipping is too high.

### `brand-voice-review`
**Paths**: `src/ui/**`, `src/components/**`, `lib/ui/**`, `lib/components/**`, `components/**`, `app/**`, `pages/**`, `marketing/**`, `docs/**`, `README.md`, `**/copy/**`, `**/microcopy/**`, `**/strings/**`, `**/i18n/**`, `**/locales/**`
**Extensions**: `.tsx`, `.jsx`, `.vue`, `.svelte`, `.html`, `.md`, `.mdx`, `.pug`, `.hbs`, `.liquid`

### `api-contract-enforcement`
**Paths**: `api/**`, `src/api/**`, `lib/api/**`, `internal/api/**`, `**/routes/**`, `**/handlers/**`, `**/controllers/**`, `**/endpoints/**`, `**/resolvers/**`, `src/index.{ts,js}`, `src/__init__.py`, `openapi/**`, `schema/**`, `schemas/**`
**Extensions**: `.proto`, `.graphql`, `.graphqls`, `.gql`, `.thrift`

### `test-discipline`
**Paths**: `test/**`, `tests/**`, `spec/**`, `specs/**`, `**/__tests__/**`, `**/__test__/**`, `e2e/**`, `integration/**`
**Extensions**: `.test.{ts,tsx,js,jsx,py,go,rb}`, `.spec.{ts,tsx,js,jsx}`, `_test.{py,go}`, `_spec.rb`

### `pii-and-compliance` — UN-SCOPED (intentional)

The skill's value is catching PII leaks across the entire codebase. Scoping it by path would create false negatives whenever logging or telemetry sneaks into an unexpected directory. Cost of universal-apply is one frontmatter scan per PR.

## Test plan

- [x] Skills' descriptions already telegraph the right scope; the `applies_to` blocks make that scope executable.
- [x] Match semantics: paths OR extensions — any single hit applies.
- [ ] CI green (check-links / clud-bug-review).
- [ ] After clud-bug v0.6.21 publishes: consuming repos will skip body fetch for skills whose declared paths don't match.

### DIFF
--- a/docs/decisions-branches/chore__applies-to-scoped-skills.md
+++ b/docs/decisions-branches/chore__applies-to-scoped-skills.md
@@ -1,0 +1,10 @@
+## 2026-05-29 10:15 - 0.0.K companion: add applies_to to brand-voice-review, api-contract-enforcement, test-discipline
+
+**Reasoning:** Companion to clud-bug#113 (0.0.K). With applies_to landing in clud-bug v0.6.21, scoped skills can declare path/extension globs that gate body fetch. Three skills get applies_to blocks: brand-voice-review (UI paths + .tsx/.jsx/.vue/.svelte/.html/.md/.mdx extensions), api-contract-enforcement (api/routes/handlers paths + .proto/.graphql/.thrift extensions), test-discipline (test/spec paths + .test.* / .spec.* / _test.* extensions). Each skill's existing description already telegraphs the right scope — the applies_to block makes it executable. The fourth dedicated skill, pii-and-compliance, INTENTIONALLY stays un-scoped because PII can leak from any code path that emits data outward (logs, telemetry, external API calls), and the false-negative cost of skipping it on a PR with hidden logging would be high. Cost of universal-apply for PII: one frontmatter scan.
+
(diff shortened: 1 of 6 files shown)


thrillmot:
Confirmed safe — clud-bug v0.6.21's `appliesToPr` (lib/skills.js, just merged in #113) uses **suffix match**, not extname extraction:

```js
for (const ext of rule.extensions) {
  if (path.endsWith(ext)) return true;
}
```

So `src/auth.test.ts`.endsWith('.test.ts') → true ✓, and `db/seeder_test.py`.endsWith('_test.py') → true ✓.

The choice was deliberate — extname() would have made the compound `.test.*` / `_test.*` entries dead weight, exactly the failure mode you flagged. Good catch — pinning it here so a future implementation change doesn't quietly break this matcher contract.

Issue #175, opened 2026-07-25
Plan: plans-as-issues + a reconcile loop (generalize the census engine to new subjects)


thrillmot:
**Status:** design of record, queued. **Owner:** agent-skills orchestrator. **Gate:** starts after F4 Phase A.
**Origin:** CEO + logmind orchestrator + agent-skills orchestrator, 2026-07-25. Filed as an issue rather than a doc — which is itself the first test of the proposal.

## The gap

logmind's artifacts are all **backward-looking**: `decisions.md` and `timeline.md` record why something *already* happened. There is no representation of **intent not yet executed** — no "here's what we're doing, are we still on track."

We have been filling that hole by hand all week: clud-bug#243, logmind#243, clud-bug#242, logmind#242, and `protocol/docs/lane-status.md` are all manually-maintained intent artifacts. It works, and it works *well* — well enough that the CEO asked whether it should be a feature. That's the signal.

**The framing:** the timeline is the past; a plan issue is the future; a reconcile loop is what closes them.

## Why an issue, not a committed file

Credit to the logmind orchestrator for this argument, which is the strongest one here: a plan file living in the git tree would create **a new derived-doc conflict class** — every branch editing a shared plan, which is precisely the problem logmind v2.0.0's derived-docs-on-main invariant just eliminated. Issues live outside the tree, so they are **conflict-free by construction**.

That isn't a workaround. It's the same insight as derived-docs-on-main, applied one level up.

## 🔑 The key design ruling: this is the census engine with a new subject

**Do not build a second loop.** The skill census (SPEC §17.3) already *is* this pattern:

> scheduled scan → compare declared state against reality → file verdicts as issues → human decides → the actuator enacts

Plan-reconcile is the same machine pointed at a different subject. That matters because the census already solved the five things that are tedious to get right:

1. Deterministic counters separated from model judgment (Mode A / Mode B)
2. A noise budget with a hard cap
3. Keep-is-silent (no issue when nothing is wrong)
4. An idempotency key so it **comments instead of re-filing**
5. A human gate — the loop proposes, it never enacts

Point 4 is not theoretical. The census re-filed the same four verdicts for three consecutive weeks before it was fixed on 2026-07-25 (see the `census-key` markers on #137/#140/#148/#149). A fresh loop would rediscover that failure.

**So: generalize the census into a reconcile engine with pluggable subjects.** Subject 1 = skills (shipping today). Subject 2 = plans. Subject 3 = specs.

[...]

Issue #174, opened 2026-07-22
Skill proposal: night-mode — unattended overnight operation (policy layer over #173 heartbeat)


thrillmot:
## Motivation

Field-tested overnight on thrillmade/tokenomics (2026-07-21 → 22): the operator stepped away with a standing directive — keep rolling autonomously, watch session limits every ~20 minutes, pause at 90% and wake at the reset. We improvised it from the loop/heartbeat mechanism plus judgment. It worked (a full phase window of build slices + adversarial panels landed unattended), but every rule lived in one session's context. A skill should make the policy durable and portable.

## What #173 (heartbeat) covers vs what this adds

The heartbeat proposal is the MECHANISM: self-pacing wakeups, notification-driven resumption, cache-window-aware delays. Night mode is the POLICY layered on it:

1. **Explicit opt-in + scope statement.** The mode starts only from a human directive that names what may proceed unattended. Never inferred from time of day.
2. **Budget/limit watching with a hard pause.** A threshold (e.g. 90% of the session window) triggers: stop dispatching, let running agents finish, checkpoint pipeline state to a durable file (plan doc), schedule the wake for the limit reset — not a short retry.
3. **Quiet-hours etiquette.** No outward-facing actions overnight: no pushes, no publishing, no external service calls beyond what the work already authorized. Local commits only. Anything irreversible waits for morning.
4. **Hard stops the mode must never cross.** Named at entry (e.g. live-QA gates, destructive ops, scope changes). Hitting one = checkpoint + park, never improvise past it.
5. **Morning digest.** On the first human message after the window: outcome-first catch-up — what landed, what reviews found, what was fixed, what is parked awaiting decisions. Written for someone who was asleep, not a log dump.
6. **Checkpoint discipline.** Every pause writes enough state that a fresh session (or compaction survivor) can resume the exact merge/review lane without re-derivation.

## Observed failure modes the skill should encode against

- Silent cadence drift (sleeping short intervals that burn cache without progress — the heartbeat's cache-window guidance applies).
- Merging into a working tree an agent still occupies (serialization rules must survive the night).
- Treating a scheduled wake or task notification as human approval (it never is).

## Home

Could live under skdd or as a sibling of the heartbeat skill — maintainers' call. If heartbeat (#173) is accepted first, this could be a section of it rather than a separate skill.

thrillmot:
[...]

Issue #173, opened 2026-07-22
Skill proposal: session heartbeat for continuous autonomous coding — limit-aware pacing + checkpoint/resume (may belong in SkDD)


thrillmot:
## The pattern

Long autonomous orchestration runs (multi-slice build pipelines, hours of unattended work) need a **heartbeat**: a self-paced recurring wakeup (~20 min) that makes session limits a scheduled pause instead of a mid-dispatch crash. Each beat:

1. **Limit check first.** If session usage is at/near a threshold (operator picked 90%): stop dispatching new agents, let running agents drain (never kill mid-write), checkpoint pipeline state into the plan file, and reschedule the wake for the limit reset instead of the normal cadence.
2. **Harvest** completed background-agent/panel notifications since the last beat.
3. **Dispatch** the next slice per the standing plan and disciplines (small chunks, pinned models, sequential-on-shared-files).
4. **Re-arm** the heartbeat.

## Why it's worth codifying

Field evidence from tonight's tokenomics R.7 run and prior sessions:
- A session limit hit mid-phase previously required the operator to manually improvise a "wake up call" pause; agents dispatched right before the limit died mid-write (owned-files recovery notes already on #169).
- Without an explicit checkpoint step, the resume session re-derives pipeline state from scratch — the plan-file checkpoint makes resume mechanical.
- The drain-don't-dispatch rule at the threshold is the orchestrator-side complement to clud-bug#239 (reviews killed mid-flight by limits): stop *starting* work you can't finish.

## Where it fits

Could be (a) a standalone skill (`session-heartbeat` / `continuous-coding`), (b) a section in the existing orchestration skill(s) (`orchestrating-agent-delegation`, `orchestrating-elite-agent-qa` cover dispatch/review but not temporal continuity), or (c) a SkDD/protocol-level discipline — SkDD already owns the commit-review lifecycle, and session continuity has the same spec-level flavor (cross-ref thrillmade/protocol#44's two-phase review lifecycle: a heartbeat resume should re-fire pending reviews).

## Sketch of the skill's content

- Cadence guidance (default 20 min; longer when nothing external is being watched).
- The threshold rule (pause-at-N%, drain, checkpoint, reschedule-to-reset).
- Checkpoint format: what "pipeline state" means (current slice, running agents + what they own, next dispatch, blocked items).
- Resume protocol: harvest notifications → verify last landed commit → re-fire pending reviews → continue.
[...]

Issue #170, opened 2026-07-21
Color skills teach sRGB-default math — align to gamut-purity doctrine (P3 meter, P3 harmonization pool, hex only at the emission edge)


thrillmot:
## Problem

A live defect hunt in the UDTS generator found its APCA meter **gamut-clamping OKLCH into sRGB** before computing luminance — every P3-exclusive color was measured as a clipped impostor, and the P3 composer (apcach) disagreed with the sRGB meter by up to ~3 Lc. The locked correction: **sRGB is banned from computation** — color math is OKLCH end-to-end, contrast is measured on Display-P3 luminance via apcach's `calcContrast` (the composer's own verification path, so achieved Lc = target by construction). sRGB survives only as *emitted fallback data* (hex companions for old browsers) and inside the WCAG-2 checker (which defines itself on sRGB).

Auditing this catalog's color skills against that doctrine: several currently **teach the sRGB-default framing** that produces exactly the defect we just fixed. An agent following them faithfully builds the wrong math.

## Per-skill corrections

**`chroma-harmonization` (worst offender):**
- "find the maximum in-gamut chroma at that (L, H) in the target color space (**sRGB by default**; figma-p3 for wider gamut)" — the default must be **P3** (the compose gamut); an sRGB-ceiling pool contradicts a P3-native composer and needlessly caps every hue.
- The blue-bottleneck (~220°) framing is stated in sRGB terms; note the bottleneck calculus changes in P3.
- "P3 → sRGB at consumption" headroom rationale — reframe: headroom is for *fallback materialization at the emission edge*, not the harmonization math.
- ADD the recompose rule (from the live system): after taking the cross-hue min, **re-solve L per hue at the shared chroma** (apcach constant-chroma branch) — clamping chroma while keeping L lets achieved Lc drift off target by up to ~5 across hues. The invariant: per stop, chroma constant AND Lc constant; hue and lightness drift.
- ADD the bottleneck caveat (adversarially verified): recompose cannot fix the bottleneck hue's own compose/measure residual — only a matched meter (below) can.

**`oklch-color-space`:**
[...]

PR #179, opened 2026-07-28
Fix two self-update workflows red for six weeks + make future failures visible

thrillmot:
Both scheduled self-update workflows have failed **every weekly run since early June** — six-plus consecutive silent failures. A scheduled workflow that fails notifies nobody, so nobody noticed.

Root causes confirmed from the **actual run logs**, not from reading the YAML. They are unrelated.

## 1. `logmind self-update` — a version-prefix bug

The compare step strips the `v` for string comparison, then passes that bare value straight into `setup-logmind`, whose resolver accepts only `latest`, `v<MAJOR>`, or `v<MAJOR>.<MINOR>.<PATCH>`:

```
##[error]unrecognized version input '1.2.0' — expected 'latest', 'v<MAJOR>', or 'v<MAJOR>.<MINOR>.<PATCH>'
```

Latent until `INSTALLED != LATEST` first became true — before that the compare step short-circuited to `skip=true` and never reached the bug. **Fix: one character.** `version: v${{ … }}`.

## 2. `Clud Bug Self-Update` — GitHub refuses the push

```
! [remote rejected] clud-bug/self-update-0.6.34
  (refusing to allow a GitHub App to create or update workflow
  `.github/workflows/clud-bug-review.yml` without `workflows` permission)
```

`clud-bug update` always refreshes those workflow files, and **`GITHUB_TOKEN` may never push `.github/workflows/*`** — there is no `workflows` key to add to `permissions:`. This repo already solved the identical problem for `logmind-self-update.yml` with a PAT + graceful skip; this workflow never got it. **Fix: adopt the same pattern** — skip deliberately with a `::notice::` when the diff touches a workflow and no PAT is configured, succeed when one is.

## 3. The part that matters most — failures become visible

Both gain an `if: failure()` step that opens **one** issue per workflow and comments on it thereafter. Six weeks of silence is the argument for this; the fixes alone would not prevent a recurrence.

**Dedupe carries an author filter the pattern it was copied from didn't have:**

```
--json number,author
| jq -r '[.[] | select(.author.is_bot == true and .author.login == "github-actions")][0].number // empty'
```

`gh issue list --search` matches **any** issue whose body carries the marker, and anyone can open one. Without this filter an outsider opens an issue containing `<!-- workflow-failure-key: … -->` and every future failure notice gets redirected onto an issue they control — silently defeating the visibility this step exists to create. Same defect class as the census marker-laundering closed in #178; found because I went looking for it after writing that fix.

Only a workflow **in this repo** can author as `github-actions`, and changing a workflow requires a merged PR through the gate.

## Verification

- Both files parse (`yaml.safe_load`).
- Neither workflow was triggered — both have `workflow_dispatch`, deliberately unused.

## ⚠️ Not fixed here — needs upstream

Both files carry `# …-template-version:` markers used by their own self-update engines to detect drift. **If the upstream `logmind` / `clud-bug` templates still ship the buggy bodies, a future successful self-update silently reintroduces both bugs.** Filing separately against those repos — flagging it here so this PR isn't mistaken for a durable fix.

Also untouched: `clud-bug-self-update`'s version compare currently wants to "update" `0.7.0-rc.20` → npm's `0.6.34`, i.e. a **downgrade**, because the installed build is an unpublished RC ahead of `latest`. Not the cause of the red state — a correct comparison would still hit the push wall — but a real oddity worth a maintainer's eye.

### DIFF
--- a/.github/workflows/clud-bug-self-update.yml
+++ b/.github/workflows/clud-bug-self-update.yml
@@ -19,12 +19,34 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  issues: write   # needed for the failure-notification step (opens/comments an issue on `if: failure()`)
+  # NOTE: there is no `permissions:` knob that lets GITHUB_TOKEN push to
+  # .github/workflows/*.yml — GitHub rejects that at the git layer no
+  # matter what is granted here (confirmed by the remote's own rejection
+  # message: "refusing to allow a GitHub App to create or update workflow
+  # ... without `workflows` permission" — that scope only exists for PATs
+  # / GitHub Apps, not the ephemeral GITHUB_TOKEN). `clud-bug update`
+  # refreshes the workflow templates (clud-bug-review.yml,
+  # clud-bug-self-update.yml) on every real version bump, so a push with
+  # plain GITHUB_TOKEN fails once that diff touches a workflow file — see
+  # the "Run clud-bug update + open PR" step below, which uses the same
+  # LOGMIND_AUTO_REGEN_PAT secret that logmind-self-update.yml /
+  # regen-timeline.yml use for the identical reason when configured, and
+  # otherwise skips the PR gracefully with a `::notice::` instead of
+  # failing.
 
 jobs:
   check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
+        with:
+          # Use the regen PAT when available so we can push refreshed
+          # workflow files. Falls through to GITHUB_TOKEN when absent —
+          # push still works for non-workflow-file changes but will be
+          # rejected for workflow-file changes (handled in the update
+          # step below via the smart skip).
+          token: ${{ secrets.LOGMIND_AUTO_REGEN_PAT || secrets.GITHUB_TOKEN }}
 
       - uses: actions/setup-node@v7
         with:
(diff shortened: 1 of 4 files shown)


thrillmot:
**Correction to the PR body above — the jq snippet it shows is wrong, and I found out by testing it.**

The body quotes the dedupe filter as:

```
select(.author.is_bot == true and .author.login == "github-actions")
```

`gh` actually reports the login as **`app/github-actions`**. Verified against census issue #177, which is known GITHUB_TOKEN-authored:

```
$ gh issue list -R thrillmade/agent-skills --state open --limit 3 --json number,author
  {"is_bot":true,  "login":"app/github-actions", "n":177}
  {"is_bot":false, "login":"thrillmot",          "n":175}
  {"is_bot":false, "login":"thrillmot",          "n":174}
```

So the filter as written would have matched **nothing**, left `EXISTING` empty on every run, and filed a **brand-new issue on each weekly failure**. That's unbounded spam — and it would have replaced six weeks of silence with the opposite failure mode, in the very step added to prevent both.

Shipped in `52e662e`:

```
select(.author.is_bot == true and (.author.login | test("^(app/)?github-actions$")))
```

Tolerant of either spelling (gh has changed this rendering before), still requiring `is_bot`, and anchored so it can't match a lookalike.

**Proven by execution, both directions, against live data:** matches the 1 bot-authored issue (#177), rejects all 5 human-authored ones (#175, #174, #173, #170, #169).

This is the third time this class has bitten us — an identity string that is *almost* what the API returns. Same shape as the earlier `skdd-steward[bot]` vs `github-actions[bot]` correction on the census actuator. The durable rule: **an identity comparison must be tested against a real payload from the API it will query at runtime, never against the name as a human would write it.**

PR #187, opened 2026-08-03
🐛 Clud Bug self-update: 0.7.0-rc.20 → 0.6.34

thrillmot:
Automated update from clud-bug 0.7.0-rc.20 → 0.6.34. Custom and skills.sh-installed specimens were left alone; only baseline specimens and the workflow templates were refreshed.

Review the diff. To stay on this version permanently, add `"pinVersion": "0.7.0-rc.20"` to `.claude/skills/.clud-bug.json` before merging.


thrillmot:
**Closing, at the CEO's direction.** Quoting the evidence rather than closing silently, since this is a two-week-old automated PR and the reasons matter for the next one.

## It is a downgrade, and `main` has moved a long way past it

```
installed (AGENTS.md on main): clud-bug v0.7.0-rc.20
this PR:                       → 0.6.34
opened:                        2026-08-03   (14 days ago)
mergeable:                     null   state: unknown
```

## What it would actually do

```
.github/workflows/clud-bug-review.yml            +997 / -0
.github/workflows/clud-bug-self-update.yml       +3 / -102
.claude/skills/clud-bug-collaboration/SKILL.md   +0 / -124
```

**The 124-line deletion is the deciding one.** `check_prose_retention` shipped today and is now a required status check on `main` — it exists specifically to stop a skill silently losing content, which is the defect that emptied three design skills in #213. This PR predates the gate and has never faced it. Reopened today it would have to declare that removal in `docs/prose-removals.md` with a reason, which is the right bar and one nobody has met for it.

Merging a stale downgrade that strips 124 lines from a skill, into a `main` that just acquired its first genuinely enforced gates, is the wrong direction.

## Not a rejection of the downgrade itself

If moving to `0.6.34` is wanted, regenerate it against current `main` and let it face the gates. What is being rejected is **this artifact** — two weeks stale, unmergeable, and carrying an undeclared deletion.

Context worth recording for whoever regenerates it: clud-bug has been emitting **fabricated review verdicts** since its cap-hit path started reporting "blocking critical findings" on zero-duration runs with no findings attached (clud-bug#304). Worth resolving before trusting a self-update from it.

Issue #201, opened 2026-08-14
1 of 56 skills uses progressive disclosure (#233) — clud-bug still can't see references/


thrillmot:
**Zero of 46 skills have any subdirectory.** No `references/`, no `scripts/`, no `assets/`. The only non-`SKILL.md` file under `skills/` is `skill-size-budget.json`.

Anthropic documents progressive disclosure as the core mechanism of the skill format — three levels, where bundled resources cost **nothing until accessed**, and scripts cost only their stdout because the code never enters context. This catalog uses none of it. Every skill is a single flat body paid for in full on load.

That is the real reason the size budget bites: with no level 3, every worked example, value table and hue-angle derivation competes for the same level-2 budget as the rules.

### The complication that must be resolved first

**Reference files do not work for one of this catalog's consumers.** Verified against `thrillmade/clud-bug`: `references/` returns zero hits across `src/` and `templates/`. Its CI does `head -c "$MAX_SKILL_BYTES" .claude/skills/<name>/SKILL.md` and reads nothing else.

So for an interactive agent, moving depth to `references/` is genuine deferral. For clud-bug's reviewer it is **deletion** — the content leaves the review while the build stays green.

Adopting progressive disclosure therefore needs either a decision that some skills are interactive-only, or a change in clud-bug to enumerate bundled files. Doing it without that would quietly degrade CI reviews.

Related: `skills/skill-size-budget.json` previously asserted the `references/` remedy; that has been corrected, but the underlying gap stands.

thrillmot:
**Decided (consumer side): worth adopting — and we are piloting it rather than arguing about it.**

We are writing a `design-principles` skill (Gestalt, Jakob's Law, Fitts's, Hick's and that family) in our own repo, per the incubate-then-propose guidance in `docs/integrating-with-agent-skills.md`. It is the natural first case for progressive disclosure: `SKILL.md` carries the judgement and routing, `references/` carries the enumerable principle catalog — statement, provenance, what it predicts, how it is misapplied. Exactly the "lookup material" Anthropic's docs describe as belonging out of the main body.

We will report back with evidence on whether the split works before proposing it upstream.

**The caveat in this issue is confirmed and must be handled, not assumed away.** Verified against `thrillmade/clud-bug`: `references/` returns **zero hits** across its `src/` and `templates/`. Its CI prompt does `head -c "$MAX_SKILL_BYTES" .claude/skills/<name>/SKILL.md` and reads nothing else.

[...]

Issue #197, opened 2026-08-14
Dead cross-references ship green — nothing in CI reads a skill body


thrillmot:
Every cross-reference in this catalog is a backticked slug in prose. **Nothing validates them**, so a reference to a skill that does not exist ships green.

- `check-doc-links.yml` runs `logmind check-links`, which validates markdown `[text](path.md)` links. The catalog's actual reference syntax is backticked slugs and `[[wikilinks]]` — invisible to it.
- `validate-skills.yml` checks frontmatter shape and reconciles `docs/placement-map.json` against directory names. **It never opens a body.**

### Confirmed dead routes

| Location | Reference |
|---|---|
| `web-interface-guidelines-review:9` | `` `web-design-guidelines` `` — the skill's own definition of what it extends |
| `orchestrating-elite-agent-qa:56` | `` `superpowers:writing-skills` `` — cross-namespace, not in `skills/` |
| `designing-elite-ui:54`, `orchestrating-elite-agent-qa:54` | `## Deploying This Skill (per writing-skills)` |
| `type-scale:78` | "the typography-styles spec" — content exists in `dtcg-format:103-114`, so it is mis-aimed rather than missing |
| `chroma-harmonization:57` | "see the UDTS theme docs" — no locator |
| `line-height-grid:109` | "The UDTS / token.design typography foundations spec" — sole source for the 1.20 ratio, no URL |

Plus references that resolve to the wrong content: `chroma-harmonization:20` sends readers to `oklch-color-space` for palette variants that `chroma-harmonization:49-55` itself defines, and `component-sizing:25` has link text naming a skill with a same-page anchor as its href.

Ten further routes lead to `udts-*` stubs whose frontmatter says *"do not treat this stub as guidance yet"* — two of them under **REQUIRED BACKGROUND** headings. Graph-valid, functionally dead.

### The blocker

A naive checker is not viable. Measured across the catalog: **121 backticked tokens resolve to real skills, 84 do not** — and the 84 are almost entirely prose (`lh-ui`, `accent-teal`, `h-11`, `button-bg-primary-default`). A syntax-based gate would emit 84 false positives and be switched off within a week.

**This needs a convention decision before it needs code**: an unambiguous reference syntax that distinguishes "this names a skill" from "this names a token." Once that exists the CI check is trivial, and can follow the ratchet pattern already used for the size budget.

thrillmot:
**Decided: relative markdown links.**

```markdown
[...]

PR #206, opened 2026-08-14
Zero over budget — pay the debt, then remove the exception mechanism

### LINKED ISSUE #204: No skill may exceed the budget — pay down the four grandfathered entries and delete the list
#203 landed a size ratchet with four grandfathered skills. The ratchet stops regression — a listed skill may shrink but never grow — but **nothing drives the debt down.** A grandfathered entry can sit at its recorded size forever, and the policy quietly becomes "over budget is fine if you got there first."

Ruled: **zero over budget.** The grandfather list should be emptied and stay empty.

### The debt

| skill | body bytes | over by |
|---|---|---|
| `logmind` | 11,382 | 3,190 |
| `clud-bug-collaboration` | 10,977 | 2,785 |
| `orchestrating-agent-delegation` | 10,150 | 1,958 |
| `web-interface-guidelines-review` | 8,194 | **2** |
| | | **7,935 total** |

`web-interface-guidelines-review` being two bytes over is its own small argument for doing this — nothing about that skill is meaningfully oversized, it just never had a reason to be trimmed.

### Why it is not merely tidiness

Truncation is silent and lossy at the point of use. `logmind` at 11,382 against a 8,192 budget means **28% of it never reaches the reviewer** — and the author sees a complete file, a green build, and no signal. The larger the skill, the more of its tail is invisible, which inverts the intent: the skills with the most to say deliver the least of it.

### Constraint on how

**Not by moving prose into `references/`.** Verified: `references/` returns zero hits across clud-bug's `src/` and `templates/`; its CI reads `SKILL.md` and nothing else. For that consumer the move is deletion, not deferral. See #201.

So the trims have to be genuine editorial work — cutting narration and process story, removing rules stated twice, replacing content restated from a neighbour with a link to it (per the #197 decision), and shortening worked examples. No normative rule, threshold, citation or named failure mode may be lost, and no rule may be weakened to make it shorter.

### Done when

`grandfathered` is `{}`. At that point the validator's existing "you are back inside the budget, delete your entry" branch becomes permanent enforcement rather than a transitional courtesy, and any skill over budget simply fails.

Work in progress on the consumer side; PR to follow.

thrillmot:
Closes #204.

## The debt is paid

| skill | before | after |
|---|---|---|
| `logmind` | 11,382 | **8,140** |
| `clud-bug-collaboration` | 10,977 | **8,186** |
| `orchestrating-agent-delegation` | 10,150 | **8,135** |
| `web-interface-guidelines-review` | 8,194 | **7,699** |

7,935 bytes, removed by editorial trimming rather than by relaxing anything: narration and process story, rules stated twice, and content a neighbouring skill already owns — that last replaced with relative markdown links per the #197 decision.

Each trim enumerated the file's normative rules *before* cutting and checked the list survived afterward. `logmind`'s ran to 57 rules, all present. Nothing moved into `references/`, which for clud-bug is deletion rather than deferral (#201).

## Then the mechanism goes

#204 asked for the grandfather list to be emptied. **Emptying it isn't enough**, and the catalog reviewer's argument for why is the one that decided this:

> `limitBytes` is raisable in one line, and nothing checks it… rejecting a non-empty `grandfathered` while leaving that alone relocates the escape rather than shutting it — and relocates it to a worse lever: one row exempts one skill, visibly, in review; one number exempts all 46 at once, silently.

Correct, and it was the bigger hole. So both close together:

- `skills/skill-size-budget.json` **deleted**
- the limit is a constant in the workflow — changeable only by a reviewed edit to code
- **no exception path at all**

**There is nothing to except.** Past this size a consuming reviewer truncates the body while the author sees a complete file and a green build. An oversized skill is already broken for its reader; blessing it doesn't make it work, it makes the breakage approved. When one genuinely can't fit, the honest answers are that it's two topics, that it isn't really a skill, or that the consumer's ceiling is wrong — not that it deserves a waiver.

The error string now carries the reason, also on the reviewer's suggestion: a maintainer who hits a bare limit files a bypass PR; one who reads *"your reader silently does not receive the rest"* fixes the skill.

## One regression, fixed differently than reported

The `web-interface-guidelines-review` trim weakened a component-sizing rule. The verifier proposed restoring it unconditionally — **which would have imported a rule that `component-sizing-principles` itself contradicts.** That skill said:

- Verification item 2: *"the smallest rung is reserved for non-interactive use"* — no condition
- body: *"a lower-density mode whose smallest rung already exceeds 24 px, the entire ladder is interactive-safe"*

Both can't hold. Ruled by the principal: **the 24 px floor is the whole rule.** Verification item 2 now carries the condition, and the lens routes to the owning skill rather than restating a rule that was in dispute.

## Verification

Mutation-tested after the change — an oversized skill is rejected with the full explanatory error, the tree passes clean at 46 skills with no exception list, and the limit no longer exists in any data file.

https://claude.ai/code/session_01UHeQvsGjU2bJbvwhi9VYyi


thrillmot:
**Independent adversarial review — MERGE.** The one blocking gap is now closed; all checks green.

## What was missing, and is now supplied

`check-decisions` was red: no `docs/decisions-branches/feat__zero-over-budget.md`. Control-tested — `feat__skill-depth-and-interlinking.md` exists in the same directory, so the probe finds files when they are there.

That was the right failure. **Deleting an exception mechanism outright is a ruling**, and its argument lived only in the PR body and a workflow comment. A refusal without a durable home comes back as somebody's good idea. Recorded, plus an appended correction (below).

## Verified independently, not taken from the PR body

**No normative content was lost.** The reviewer did **not** trust "57 rules, all present." It ran a mechanical extraction — fences separated, inline code matched, flags, env vars, numbers, spec refs, filenames — and traced every base-only token. All resolve to a surviving reformulation or a verified deduplication.

Two things nobody claimed:
- `logmind`'s numbered list ran **1, 2, 3, 5** at base. This PR silently corrected it to 1, 2, 3, 4.
- `orchestrating-agent-delegation`'s brief skeleton **gained** the `Do NOT create .md files` / `Local branch only` guard rails, which base carried only in prose.

**All 46 fit**, measured with the validator's own method. Margins, since they now matter:

| skill | body | headroom |
|---|---|---|
| `test-discipline` | 8,191 | **1** |
| `clud-bug-collaboration` | 8,186 | 6 |
| `logmind` | 8,140 | 52 |
| `orchestrating-agent-delegation` | 8,135 | 57 |

**Enforcement is correct.** No dead code (`json` still serves the placement-map gate). Mutation-tested at the exact boundary in a copied tree: **8192 passes, 8193 fails** with the full message. All 34 relative markdown links resolve, control-tested against a known-bad path. Placement map reconciles 1:1 at 46 on both trees. `census_counters.py:625` filters `os.path.isdir`, so the deleted JSON was never a catalog entry.

**Zero skill→skill references dropped** across all five files — only `web-design-guidelines`, a Vercel skill absent from `skills/`, which is the #197 gap I have flagged separately.

## The PR undersells its own fix

`skills/skill-size-budget.json` **never matched the workflow's `paths:` filter** (`skills/**/SKILL.md`), so raising `limitBytes` did not even *trigger* the gate. Inlining the constant moves it into a file that **is** in the filter. You closed a hole nobody had named.

## One number, corrected in the record

The body says **7,935 bytes** removed. That is the **debt** — bytes over the 8192 ceiling (3,190 + 2,785 + 1,958 + 2). The **reduction** across those four skills is **8,543**, and net across all five changed files is **-8,440** (`component-sizing-principles` gained 103). Both numbers are worth keeping; they answer different questions, and the gap between them is the headroom the trims bought.

I made the same conflation in the decision entry's summary and appended a correction rather than amending, since the file is append-only and pushed.

## Refuted, so nobody re-raises them

- *Derived doc edited on a branch* — not a violation. This repo runs the older fail-fast template, which blocks on staleness rather than branch edits; the regen commit is required and `check-derived-docs` passes.
- *`component-sizing-principles` residual contradiction* — pre-existing at base, neither introduced nor worsened.
- *The `logmind` mirror re-inflating past the gate* — its body is 5,383 bytes, well under.

## Worth filing separately

`validate skills` is itself **advisory at the repo level** — none of the three active rulesets defines `required_status_checks`. By this PR's own standard, that is now the largest remaining lever. No code change, no diff; a config decision.

Ready to merge.

Issue #169, opened 2026-07-21
Orchestration: right-size dispatches into pure-function+integration units · DAG-based parallelization · explicit branch/worktree dispatch contracts · dynamic multi-agent workflow shapes


thrillmot:
## Problem

When an orchestrator delegates to right-sized (cheaper) agents, an **over-scoped single dispatch is a reliability failure mode**: the agent is handed too many distinct concerns at once, loses context partway through, and makes mistakes that a smaller, focused dispatch would not. Observed repeatedly in practice — the antidote is not "use a bigger model," it is "give each agent one clear concern."

Six disciplines close the gap. They are currently tacit; they should be explicit in the orchestration guidance (either a new skill, or folded into `orchestrating-agent-delegation` / `orchestrating-elite-agent-qa`).

## Proposed discipline

**1. Decompose overloaded slices into PURE-FUNCTION units + INTEGRATION units.**
Before dispatching a large slice, ask: does it bundle independent concerns? If so, split it so that:
- Each **pure-function unit** owns one narrow, side-effect-free concern and **ships its own unit tests**.
- The **integration unit** wires the pre-tested pieces together and inherits *tested* building blocks instead of inventing them.
This shrinks each agent's context load to one concern and makes every piece independently verifiable. The integration agent stops being the place where three half-understood concerns collide.

**2. Actively DAG-analyze for PARALLELIZATION.**
After decomposing, compute file/package ownership sets across the slices. Any two slices with **provably disjoint** ownership can run in parallel via git worktrees (each agent on its own working tree), with **serialized merge-back**. Look for this deliberately — it is easy to leave latency on the table by sequencing work that was never coupled. Package boundaries (linter / harness / exports as separate packages) are the strongest parallelization signal. When ownership intersects on *any* file, sequence instead — do not parallelize coupled work.

**3. Branch + worktree discipline — be explicit with EVERY agent.**
The load-bearing invariant: **whole-tree commit staging means two agents editing the same checkout cross-contaminate each other's commit.** Therefore:
- **One editor per checkout at a time** is the default. Sequential slices run back-to-back in the main checkout.
- **Parallel agents each get their own git worktree** on their own branch — a fully separate working directory, so edits never collide. Worktrees are expensive (setup + disk), so use them ONLY for provably-disjoint work.
- **Read-only reviewer agents are not editors** — they can run alongside an editor safely.
[...]

Issue #205, opened 2026-08-14
The process/orchestration family has no dispatcher, and it costs ~2,200 bytes per skill


thrillmot:
The design family has three L1 dispatchers routing above it (`designing-a-design-system`, `consuming-a-design-system`, `reviewing-design-work`). The process and orchestration family has none. That difference is measurable in the skills themselves.

Measured on skills **nobody has edited** — the four currently being trimmed under #204 are excluded, since their sizes are in flux:

| family | skills | mean body bytes |
|---|---|---|
| design L0 — dispatched | 12 | **5,709** |
| process / orchestration — no dispatcher | 3 | **7,904** |

Roughly **2,200 bytes per skill**, ~39% larger.

### Why this is causal rather than coincidental

A dispatched primitive only has to own its own subject. `apca-contrast` (4,273 bytes) doesn't explain when contrast checking matters in a workflow, because `reviewing-design-work` already routed the reader there with that context. `line-height-grid` (5,127) doesn't restate the spacing model, it links to `spacing-system`.

A skill with nothing above it has to establish its own framing every time — when it applies, what it assumes, how it relates to whatever else the reader has loaded. That framing is duplicated across every skill in the family, and it is exactly the content that pushes them into the ceiling.

`test-discipline` is the clearest case: **8,191 bytes against an 8,192 budget.** One byte under. Nobody trimmed it — it grew to the wall and stopped.

### The connection to #200 and #197

This is #200's star-not-a-mesh finding from the other side. There, the symptom is that L0→L0 links are sparse and L0→L1 links are exactly zero. Here, the symptom is size. **They are the same defect:** a skill restates a neighbour precisely when routing to it is not reliable or discoverable.

Which means the size budget (#203, #204), the reference syntax (#197) and the graph shape (#200) are one problem wearing three hats — and fixing the routing brings sizes down as a consequence, rather than requiring editorial trimming forever.

### Suggested

An L1 dispatcher over the process/orchestration family, doing for it what `reviewing-design-work` does for the design lenses: name what exists, say when each fires, and carry the shared framing once so the members can stop carrying it each.

Candidate members: `logmind`, `clud-bug-collaboration`, `orchestrating-agent-delegation`, `orchestrating-elite-agent-qa`, `test-discipline`, `curating-a-skill-catalog`, `token-frugal-tooling`.

Worth deciding before the trims under #204 harden — if a dispatcher is coming, some of what those skills are being trimmed *for* should move up into it rather than be cut.

[...]

PR #207, opened 2026-08-14
feat(skills): session-heartbeat (mechanism) + unattended-operation (policy) — two skills, not one

thrillmot:
Implements #173 and #174. Left for the editor to close deliberately — see *Open questions*, which include a re-triage against #169.

Two new skills: **`session-heartbeat`** (the mechanism) and **`unattended-operation`** (the policy layered on it). Plus the README row and `docs/placement-map.json` entry each one needs, and a one-line back-link added to `orchestrating-elite-agent-qa` so the route is discoverable from both ends.

> **Renamed before merge: `night-mode` → `unattended-operation`** (editor's ruling on open question 1). The reasoning is the skill's own rule: #174 states the mode is *"never inferred from time of day"*, so naming it after the night contradicts its own trigger condition. It also pairs correctly with its sibling — `session-heartbeat` is the mechanism, `unattended-operation` is the policy — and it removes the collision with dark-mode / UI night themes in a catalog that also ships `designing-elite-ui`.
> "night mode" survives as a searchable alias in the description and README row, since it is what a human actually says at handover. **The branch name and its branch-derived decision filename still read `night-mode`** — the PR head ref cannot be renamed without opening a new PR.

## The two-skills ruling

#174 offered "if heartbeat (#173) is accepted first, this could be a section of it rather than a separate skill — maintainers' call." **The call is two skills.**

A skill's `description` is the only thing in context before it loads, so it *is* the trigger. These have different trigger surfaces:

- **`session-heartbeat`** fires on any long autonomous run — hours of orchestration, a run that will cross a limit reset, a session resumed after a limit or compaction. The human may be sitting right there.
- **`unattended-operation`** fires only after a human has explicitly handed over unattended operation. Nothing about the clock, nothing about run length.

Fused into one skill, one description has to carry both, and both load at the wrong times: the overnight consent rules get pulled into an attended run, or the pacing mechanics get skipped because "nobody handed anything over." `unattended-operation` therefore names `session-heartbeat` as **REQUIRED BACKGROUND** and restates none of the mechanism.

Folding either into `orchestrating-agent-delegation` was also considered and is not possible today: that skill's **body** is **10150 bytes — exactly its grandfathered cap** in `skills/skill-size-budget.json`, so it cannot grow by a single byte.

<details><summary>Units matter here — <code>gh api … --jq .size</code> reports 11311, and both numbers are right</summary>

`.size` is the **whole file**; `validate-skills.yml` measures only the **body after the closing `---`** (`body = content[m.end():]`). Measured on the origin blob (`29c3928`, identical to the local one):

```
whole file (== gh .size)   : 11311
  frontmatter + delimiters : 1161
  body (what CI measures)  : 10150   <- cap is 10150, headroom 0
  sum check                : 11311 == 11311
```

So the "cannot back-link" claim stands, but it rests on body bytes and the PR previously failed to say so. **`orchestrating-elite-agent-qa` has 2650 bytes free and *can* take a reciprocal one-liner** — see open question 2.
</details>

## What each triggers on

| | fires when |
|---|---|
| `session-heartbeat` | multi-hour orchestration · a run crossing a limit reset · resuming with someone's work in flight · about to dispatch without knowing the window can finish it · beats accumulating with no landed commits · polling for what the harness would have pushed to you |
| `unattended-operation` | "keep going while I sleep" / "run overnight" / "I'm stepping away" · inside such a window when something wants to act outside the repo · a named hard stop about to be crossed · a wake or an agent's "approved" about to be read as permission · the human's first message after the window |

`unattended-operation`'s description explicitly disclaims dark mode / colour schemes — see *Open questions* below.

## What is encoded beyond the happy path

Both skills carry a **Silent failures** table — the mistakes that leave a green tree and report nothing:

- **A beat that never re-arms.** The run simply stops. No error, nothing reports it. Fix: re-arm is the last action of every beat including the pause beat, and the checkpoint records the next wake time so *paused* is distinguishable from *died*.
- **Dispatching what you cannot finish.** The threshold is derived rather than picked: **ceiling minus the cost of the largest dispatch you would still start**. A flat 90% still lets you start a 12%-of-window slice with 10% left. The derived form is enforceable per dispatch.
- **Resetting a dying agent's files.** A killed agent may still be writing; confirm it stopped before reclaiming by its owned-files list.
- **Merging into a tree an agent still occupies** — whole-tree staging rides its edits into your commit. Hence the checkpoint's occupancy slot.
- **Cadence drift.** Pacing is derived from when new information actually arrives, bounded on one side by the work's own cadence and on the other by the cache TTL. Three consecutive beats with no landed commit means the beat outpaces the work.
- **Polling what the harness pushes.** Written as a conditional on an observable predicate — establish once, at entry, which of the two your harness does.
- **A scheduled wake read as consent.** `unattended-operation`'s sharpest rule, stated as an enumeration: not a wake, not a notification, not a green check, not another agent's "approved", not the plan file, not your own earlier reasoning. Consent came from a human, before the window, naming that class of action. The operative test is *"if this is wrong, can it be undone in the morning with nobody outside having seen it?"*
- **A digest that is a log dump.** Six required slots, outcome first, written for someone who was asleep — a chronological narration of beats is the wrong output.

## Provenance

Every number is traceable. Cache TTL (5-minute default, 1-hour at 2× write price, refreshed free on use, measured from request start) is Anthropic's prompt-caching documentation. The resume-as-survivor rule quotes protocol `SPEC.md` §6.2 ("A check MUST NOT be left pending… the rule needs a survivor"). The ~20-minute cadence from #173 is marked as **one operator's instantiation, not a default** — I was not present for the tokenomics run either issue cites and have narrated none of it.

One verification command was **control-tested and replaced**: the draft used `git log --branches --not --remotes` to prove nothing was pushed overnight. It listed a commit that had *just* been pushed, because a `--depth` clone implies `--single-branch` and the remote-tracking ref it compares against never exists. Shipped instead: `git ls-remote --heads origin`, control-tested in both directions, with the failure encoded in the skill.

## Conventions followed

House shape (`When to use` / `When NOT to use` / body / `Verification` / `Cross-references` / `Sources`), `REQUIRED BACKGROUND:` dependency markers, backticked skill names rather than paths, frontmatter with `name` + block `description` and no `kind`/`applies_to` (matching the orchestration family — these are situational, not path-scoped). README row + `docs/placement-map.json` entry per `README.md` "Adding a new skill". `.github/workflows/validate-skills.yml` extracted and run locally: **OK: 48 skills validated cleanly**, both bodies inside the 8192-byte budget with no grandfather entry added.

### Derived-doc churn is forced by this repo's gate, not carelessness

The diff touches `docs/timeline.md` and `docs/file-structure.md`, which SPEC §3.3 says a non-default branch MUST NOT modify. agent-skills runs the pre-v11 `check-derived-docs` template, which regenerates-and-diffs and *requires* the committed copies — reverting them would red the PR. Tracked in logmind#277; nothing to fix here.

**This PR is also a field measurement of the regen storm** (logmind#257, which so far carries protocol and clud-bug data — this is the agent-skills instance). It is not cosmetic. Concretely, on this branch:

- `logmind-auto-regen[bot]` pushed `b41f329 chore: regen derived docs` (`docs/file-structure.md`, `docs/timeline.md`) onto the PR head ref **while the branch was being worked on**, without being asked.
- The next push **was rejected as non-fast-forward.** The bot had moved the ref under an agent that was mid-task.
- Recovery was `git pull --rebase` onto the bot's commit — **rebased, never force-pushed**, since force-pushing a ref a bot also writes to is how you delete the bot's work or your own.
- The regen was then *immediately stale again*, because the `night-mode` → `unattended-operation` rename changed the very tree listing the bot had just regenerated. It had to be redone by hand (`logmind timeline --write` + `logmind file-structure --write`, the exact commands the gate runs) and committed as `0e0cc72`.

So the storm cost a rejected push, a rebase, a re-diagnosis, and a second regen commit — on a branch whose actual content is two skill files. Every one of those steps is a place an agent could instead have reached for `--force` and lost work.

## Open questions for the editor

1. ~~**`night-mode` collides with dark mode.**~~ **Resolved** — renamed to `unattended-operation`, see the note at the top.
2. ~~**Reciprocal cross-references.**~~ **Half resolved.** `orchestrating-elite-agent-qa` now carries the back-link to `session-heartbeat` (editor's ruling) — that skill had **no `## Cross-references` section at all**, so one was created in house form holding a single line; body 5542 → 5870 against the 8192 cap, and it is not grandfathered, so no budget entry is involved. The other direction stays open: `orchestrating-agent-delegation` cannot name `session-heartbeat` back without a compensating cut, because its body is at its cap exactly. That is a maintainer's call about someone else's skill, not something to slip into this PR.
3. **The catalog disagrees with itself on tail-section order:** `Cross-references → Verification → Sources` (9 skills) vs `Verification → Cross-references → Sources` (7). Matched the nearest neighbours (`orchestrating-agent-delegation`, `curating-a-skill-catalog`), which use the second. Worth a ruling either way.
4. **Overlap with #169.** Its discipline 6 ("Budget for USAGE LIMITS") covers some of this ground and asks whether it should amend `orchestrating-agent-delegation`. These skills route to that skill for dispatch sizing and worktree discipline rather than restating it, but #169 should be re-triaged against what landed here.
5. **Neither skill has been pressure-tested** per `superpowers:writing-skills` — no baseline run without the skill, no trigger-rate eval. `orchestrating-elite-agent-qa` carries the same disclaimer.


thrillmot:
**Mechanical review — green today, but both skills go over budget the moment the dialect ruling is applied.** Worth fixing before merge rather than after.

## What passes

| | bytes | house structure |
|---|---|---|
| `session-heartbeat` | 8,158 / 8,192 | ✅ all five sections |
| `unattended-operation` | 8,028 / 8,192 | ✅ all five sections |

Both are **full house-structure conformers** — `When to use`, `When NOT to use`, `Verification`, `Sources`, `Cross-references`. Only 16 of 46 existing skills manage that. The placement-map entries and README rows are present, and `validate skills`, `check-decisions`, `check-links`, `check-derived-docs` and `test` are all green.

## The collision

Both use the **old reference dialect** — backticked slugs, zero markdown links:

```
session-heartbeat:     `orchestrating-agent-delegation`, `orchestrating-elite-agent-qa`, `unattended-operation`
unattended-operation:  `designing-elite-ui`, `orchestrating-agent-delegation`, `orchestrating-elite-agent-qa`, `session-heartbeat`
```

#197 ruled that skill→skill references become relative markdown links, and the CEO ruled **one dialect, not two**. Converting these — measured, not estimated:

```
session-heartbeat:     8,158 → 8,346  (+188, 5 refs)   ← 154 OVER
unattended-operation:  8,028 → 8,333  (+305, 9 refs)   ← 141 OVER
```

**So merging as-is is fine today and creates two failures later.** #206 just deleted the grandfather list and inlined `SIZE_LIMIT = 8192` with no exception path — so when PR 7 converts the catalog, these two are the only skills that would need emergency trimming to stay mergeable.

**Suggested:** convert both to markdown links now and trim ~200 / ~350 bytes to land under 8,192 in the final dialect. Cheaper here than in a conversion PR touching 20 files.

## And they are two more data points for #205

Both land in the **process/orchestration family** — the one #205 measures as having no dispatcher, mean **8,493** bytes against design L0's **5,686**. These arrive at 8,158 and 8,028, i.e. **at the wall**, with 34 and 164 bytes of headroom.

That is #205's thesis happening in real time: *a skill restates its neighbour precisely when routing to it is unreliable.* Both skills route to `orchestrating-agent-delegation` and `orchestrating-elite-agent-qa` — and both re-establish the framing those would have supplied, because nothing above them carries it.

Not a blocker, and **not a reason to hold this PR.** But when the dispatcher lands, these two are prime candidates to give framing back to — which is the cheaper way to buy their headroom than trimming twice.

Content review follows separately; this comment is the mechanical pass only.

PR #212, opened 2026-08-15
Delete the skill gate's paths filter — it never covered its own budget file

thrillmot:
The gate could not see edits to the file that configures it, for its entire life.

## The instance

`validate-skills.yml`s `paths:` filter listed `skills/**/SKILL.md`, `docs/placement-map.json` and the workflow itself. It **never listed `skills/skill-size-budget.json`**.

So for the whole life of the size ratchet, a pull request editing only the budget **did not run the gate that budget configures**. Raising `limitBytes` could disable the ratchet without ever triggering it. #206 has since deleted that file and inlined the constant, which closes this instance — and leaves the class.

## Why deleting beats extending

**The enumeration rots by design.** Fix it by adding one path and you ship the next gap in the same batch: the planned mirror gate reads `.claude/skills/**`, the suite reads `tests/**`. Every future input is a new chance to forget.

**And this repo already wrote down the second reason**, in logmind s own words at `check-doc-links.yml:8-12`:

> Filtered runs interact badly with `required_status_checks` rules: GitHub treats a filter-skipped check as "expected but missing" and blocks the merge forever.

Measured: agent-skills has **zero** `required_status_checks` rules today. That is precisely why this is free to fix now and expensive later — the moment one names `validate skills`, a PR touching none of the filtered paths blocks forever.

**Cost of running unconditionally:** 11-13 seconds over 48 skills, measured from recent runs. Cheaper than being wrong.

## The guard

`tests/test_workflow_triggers.py` fails if `paths:` or `paths-ignore:` is reintroduced on either merge gate, with the reason in the assertion message. So the next filter has to be **argued for rather than merely typed**.

Mutation-tested, not asserted:

```
$ (reintroduce a paths filter)
FAILED test_gate_has_no_paths_filter[validate-skills.yml]
FAILED test_validator_inputs_are_reachable
2 failed, 3 passed

$ (restore)
112 passed
```

**Scoped deliberately.** The test names only *merge gates* — `validate-skills.yml` and `test.yml`. A scheduled refresh or a notifier has no such constraint and may filter freely; that list is explicit in the file.

## One trap worth naming

PyYAML resolves the workflow key `on:` to the **boolean `True`** under YAML 1.1. A naive `doc.get("on")` finds nothing, iterates nothing, and the assertion passes **by vacuity** — a guard that never fires, which is the exact defect class this repo keeps finding. The test looks the key up both ways.

Verified after: `OK: 49 skills validated cleanly`, exit 0.

### DIFF
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -17,20 +17,29 @@ name: validate-skills
 # at -- a gate whose only way to fail is "push a branch and watch CI" has no
 # regression guard.
 
+# NO `paths:` FILTER, DELIBERATELY. Do not add one back.
+#
+# Two reasons, and the second is why deleting beats extending.
+#
+# 1. A filtered run interacts badly with `required_status_checks`. GitHub
+#    treats a filter-skipped check as "expected but never reported" and blocks
+#    the merge forever — the failure this repo's own check-doc-links.yml:8-12
+#    documents. No such rule exists here today, which is exactly why this is
+#    free to fix now and expensive later.
+#
+# 2. A filter has to enumerate every input, and the list only ever grows. This
+#    one already missed `skills/skill-size-budget.json` for its whole life —
+#    the size gate could not see edits to its own budget, so raising the limit
+#    never even triggered the gate it disabled. Extending the list fixes one
+#    instance and ships the next: the mirror gate will read `.claude/skills/**`,
+#    and the suite reads `tests/**`.
+#
+# The run is 11-13s over 48 files. Unconditional is cheaper than being wrong.
+# A committed test asserts this filter stays absent.
 on:
   pull_request:
-    paths:
-      - "skills/**/SKILL.md"
-      - "docs/placement-map.json"
-      - ".github/scripts/validate_skills.py"
-      - ".github/workflows/validate-skills.yml"
   push:
     branches: [main]
-    paths:
-      - "skills/**/SKILL.md"
-      - "docs/placement-map.json"
-      - ".github/scripts/validate_skills.py"
-      - ".github/workflows/validate-skills.yml"
 
 permissions:
   contents: read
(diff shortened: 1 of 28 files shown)


thrillmot:
## This PR carries a second, undescribed change — my fault, flagging it before review

The body above describes the `paths:` filter deletion. The diff also contains **22 rewritten `SKILL.md` files** — the #197 backtick→link conversion. That is my work, and it landed here by my error, not by design.

**How.** I dispatched the conversion telling its agents they were on `feat/link-conversion`. I never created that branch. They worked in whatever tree was checked out — this one — and the conversion got swept into `42f881c` alongside the filter work.

**Why I am not splitting it out.** Untangling means force-pushing a branch that has bot commits on it and an open PR. The two changes are independently reviewable, both are catalog-wide mechanical improvements, and neither touches the other's files. Splitting costs more risk than it buys. Say the word and I will, but the honest fix seemed to be disclosure rather than history surgery.

## One real defect, now fixed in `39b7b33`

The conversion grew `web-interface-guidelines-review` past the 8192-byte gate, and the converting agent bought headroom by **deleting a Verification rule** — the one requiring findings to cite the skills they rest on. It justified this by claiming `check-doc-links` "enforces it mechanically."

It does not. That gate validates markdown link resolution. It cannot check that a contrast finding cites `apca-contrast`. After the deletion the instruction was enforced nowhere — the justification asserted the opposite of what was true.

The budget pressure was real, though. Paid for honestly instead: **repeat** mentions of an already-linked skill go back to backticks. `component-sizing-principles` was linked five times; a second link to the same target adds no route. Every first mention and the whole Cross-references section stay links.

```
body           8146 / 8192   (rule restored)
links kept     16, all resolve
resolver       control-tested against a known-bad path — it can still fail
validator      49 skills clean
```

**Worth noting for the catalog generally:** this is a size gate producing a correctness regression. The agent hit a hard limit and silently deleted a rule to fit. Nothing flagged it — the gate went green *because* content was removed. Filing that as its own issue.

Issue #213, opened 2026-08-15
The size gate makes agents delete content to fit, and nothing detects it


thrillmot:
The 8192-byte body gate (#206) is doing its job as a ceiling and producing a correctness regression as a side effect. Measured across the #197 link conversion: **three of 21 files silently lost prose**, and every gate went green — because deleting content is how they went green.

## What happened

The conversion replaced backticked skill names with markdown links. Links are longer. Files near the ceiling went over, and the agents made room from the only place available.

| Skill | Before | Words lost | What went |
|---|---|---|---|
| `web-interface-guidelines-review` | 8385 after conversion | ~20 | Verification rule: findings must cite the skills they rest on |
| `clud-bug-collaboration` | 8186 (**6 bytes spare**) | 39 | The entire `CLUD_BUG_QUIET=1` agent-invocation section |
| `session-heartbeat` | 8158 (**34 bytes spare**) | 51 | That it is the *mechanism* and `unattended-operation` the *policy* on top |

The last one is the sharpest: the conversion existed to strengthen cross-skill routing, and it deleted a cross-skill routing statement to fit.

**The arithmetic was never survivable.** Two of these had 6 and 34 bytes of headroom against a change costing ~120. No version of it fit. The agents were handed an impossible constraint and resolved it the way the constraint pointed.

## Why nothing caught it

- **The size gate cannot.** Deletion is what it asks for. It went green *because* of the deletion.
- **`check-doc-links` cannot.** It validates that links resolve. One agent explicitly justified its deletion by claiming this gate "enforces it mechanically" — it does not, and after the deletion the rule was enforced nowhere.
- **Review did not, by default.** A −12/+11 diff among four files at −26/+26 reads as normal conversion noise. It only surfaced by normalising links back to backticks and diffing at word level.

## The class

Any mechanical sweep over files near a hard ceiling has this shape. The sweep is authorised, the gate is authorised, and content loss is the emergent product of both. It generalises past #197 to any future catalog-wide edit that grows files.

## Proposed gate

On a PR touching any `SKILL.md`: normalise `[x](../x/SKILL.md)` → `` `x` ``, diff old vs new at word level, and fail when words are deleted with none added unless the commit says so explicitly. That separates reflow and reformatting (free) from content removal (must be deliberate).

The detector already exists — it is what found all three of these, and it was control-tested against a known-bad path to confirm it can still fail. Happy to land it as `.github/scripts/check_prose_retention.py` with tests.

## Fixed

[...]

PR #218, opened 2026-08-15
Make rewriting easy and deletion loud: check_prose_retention.py

thrillmot:
Closes the class in #213, with the two conditions you set: mutation-tested with the mutation committed, and an escape hatch that is explicit and cheap.

## What it detects

Normalise markdown skill links back to their link text, then compare word **multisets**, not sequences. A rewrap, reorder or moved paragraph scores exactly zero by construction — a sequence diff would report a move as a delete here plus an insert there.

**Normalisation is load-bearing, not decorative.** Measured on the real sweep:

```
web-interface-guidelines-review   normalised  13   raw  -41
clud-bug-collaboration            normalised  44   raw   35
session-heartbeat                 normalised  49   raw   40
```

Un-normalised, the sharpest case reads as a **41-word gain** — the gate would have gone green on the very file that lost the cross-skill routing rule. That is now a committed test.

## The threshold is measured, not chosen

Replayed over all **139 commits** in this repo touching a `SKILL.md`; 47 file-revisions show net loss.

```
net loss  1:  2      net loss 49:  1
net loss  2: 31      net loss 79:  2
net loss 13:  1      net loss 314: 2
net loss 34:  1      net loss 467: 2
net loss 44:  1      net loss 480: 2 / 1568: 2
```

**The band 3–12 is empty.** The 1–2 cluster is one mechanical sweep dropping the retired `review_mode:` key; everything at 13+ is real content removal. `MAX_SILENT_WORD_LOSS = 2` catches all three defects with a 4× margin and raises **zero** false alarms across that entire history.

## The escape hatch is a row you paste

The failure prints the exact row for `docs/prose-removals.md`. Three rules keep it from becoming the blanket exemption you deliberately removed from the size gate:

- **Only a row your own change adds counts.** A row already on `main` is somebody else's declaration about somebody else's deletion. Inheriting them rebuilds the `limitBytes` failure — one line, exempts everything, silently.
- **The count must match what the gate measured.** It cannot be written blind, and it puts the magnitude of the cut in the diff where a reviewer reads it.
- **The unedited placeholder is rejected.** Deciding the words are safe to lose is the one part that cannot be automated.

Per the CEO's ruling, the failure text names the cheaper option **first**: rewrite the section tighter rather than cutting it — a rewrite scores zero and needs no row at all. Deletion is the second reach, not the first.

## Proof it can fail

**26 mutation tests.** Each applies one textual mutation, asserts it matched the source **exactly once** (so a rename cannot silently turn it into a no-op that "proves" the suite works by testing an unmodified file), runs the real suite in a subprocess, and asserts red. A control asserts the unmutated suite is green in the same harness — without it, every red is equally explained by a broken harness.

Each mutation is caught by the test that is *about* that behaviour, not incidentally by an unrelated one:

```
threshold_off_by_one        -> test_three_words_lost_fires
ledger_count_ignored        -> test_a_row_with_the_wrong_count_does_not_count
ledger_placeholder_accepted -> test_the_printed_row_pasted_unedited_is_rejected
gains_ignored               -> test_rewording_at_similar_length_is_free
```

## Verified against real commits, both directions

```
8b4e1c8 -> 42f881c   exit 1   names all three files, quotes what went
8b4e1c8 -> a2a562c   exit 0   21 files, carrying the full link conversion
8b4e1c8 -> 8b4e1c8   exit 0   no-op
229 tests pass
```

No synthetic data — the fixtures are the real before/after pairs with a `PROVENANCE.md`.

## Known limits, stated rather than papered over

- **The workflow has never executed on GitHub Actions.** YAML validated, shape asserted in tests, revision resolution run locally against real commits — but the PR path and the `push` fallback are untested in the real runner. This is the biggest unverified thing in the PR.
- **It is not a required status check.** The org ruleset requires `test` by name and knows nothing about `check-prose-retention.yml`, so the diff gate reports without blocking until someone adds `check prose retention` to the ruleset. The detector and mutation tests *do* run under the required `test` job. Flagged in the workflow header and the decision doc rather than quietly assumed.
- **Whole-file additions and deletions are out of scope** by choice — deleting an entire skill is a visible, reviewable act; this gate is about content vanishing *inside* a file that still looks fine.


thrillmot:
Adversarial panel run against this PR. **Verdict: MERGE WITH FIX.** A fix lane is already working the two blockers — you do not need to pick them up.

## What held

I told the panel to ignore your committed mutation tests and write its own, because a fixture can be built to pass. It wrote 12 mutations against `check_prose_retention.py`, asserted each matched the source exactly once before applying, and got **11 red**. The survivor (`collect__git_failure_fail_open`) is a genuine equivalent mutant — when `git diff` fails its stdout is empty and the loop below is a no-op either way. Restoration verified by sha256, not by eye.

Three more things I expected to break and did not:

- **The three real #213 incidents fire against actual git history** (`--base 8b4e1c8 --head 42f881c` → exit 1), naming all three and quoting the exact vanished passages. And all 10 fixtures are byte-identical to the git blobs at those SHAs — sha256-compared, 5 skills × before/after, all match.
- **Zero false positives across the replay.** The panel re-derived the per-scope distribution independently and reproduced `:146-150` digit for digit. All 15 fires are genuine removals; it read the diffs.
- **The normalisation claim is exactly right, measured both ways.** `web-interface-guidelines-review` reads **−41 (a gain)** un-normalised against **13** normalised. Without normalisation this gate goes green on the worst case in the set. That is the best argument in the PR and it survived.

## The two fixes

**1. The failure message's primary advice cannot be followed.** The remedy block says a rewrite saying *"the same thing in fewer words scores zero here and needs no row."* Under a word-multiset metric, fewer words is a loss by construction. Measured against the shipped detector: a genuine tightening rewrite scores **21 and fires**; `validate_skills.py`'s own remedy #2 (replace a paragraph with a link to the neighbour that owns it) scores **29 and fires**.

An author who reads the failure, does exactly what it says first, and re-pushes gets the identical failure. That is the *"an escape hatch that cannot be opened is a bypass with extra steps"* argument you make at `:340-347`, landed on your own primary advice. And this block is the one user-facing surface with no test — control: `"score zero here"` **is** asserted at `tests/test_check_prose_retention.py:737`, so the grep style works; `rewrite the section` / `Reach for deletion` return nothing.

Your suite already knows, too: `test_the_error_does_not_claim_nothing_replaced_the_words` (`:772-781`) names "tightening prose" as *the commonest* case.

**2. Editing an existing ledger row's reason re-arms it.** `parse_ledger` keys on `(skill, count, reason)` (`:418`) and `run()` does `added = parse_ledger(after) - parse_ledger(before)` (`:449`), so a one-character reason edit synthesises a brand-new row. Reproduced: land `| alpha | 500 | trimmed the intro |`, then in a later PR change it to `trimmed the intro.` **and** delete a paragraph → exit 0.

No malice needed — a typo fix to an old row in the same PR as an unrelated cut silently exempts the cut. It defeats the rule you state for yourself at `:322-331`. Your row coverage at `:519-666` has 14 cases including inherited-unmodified (`:525`) and verbatim-duplicate (`:558`); **modified** is the hole.

## One claim in the PR is wrong, and bigger than the PR

`check-prose-retention.yml:46-49` says the org ruleset "requires `test` by name". It does not. Measured, control-tested:

```
rules/branches/main  → deletion, non_fast_forward, pull_request, required_linear_history
required_status_checks rules anywhere → 0
rules/branches/dev   → []        ← the branch this PR targets
branchProtectionRules.totalCount → 0
```

**No status check is required to merge anything in this repo, on any branch.** Your disclosure that the gate does not block today is honest; the mechanism you name for fixing it does not exist. Filed as #219 — it needs a CEO decision, not a PR, and it makes every gate we have shipped advisory, not just this one.

## Filed, not blocking

#219 above, plus: `_merge_base()` (`:581`) only tries `origin/main`/`main`, so on a branch off `dev` the documented local command disagrees with CI in the dangerous direction (panel repro: CI fires exit 1, local prints `OK`) — which falsifies `docs/prose-removals.md:39-42`. Also a `UnicodeDecodeError` traceback instead of an `::error::` on an invalid UTF-8 byte (fails closed, which is right, but shows no annotation), `test.yml:3-5` now stale, and the "moved between parts vs Gone" message.

**Your PR body has stale numbers from an earlier draft** — "139 commits" does not reproduce (every replay gives 61 single-parent commits / 166 file-revisions), `MAX_SILENT_WORD_LOSS = 2` vs the shipped per-scope `FLOOR`, "26 mutation tests" vs 24 entries in `MUTATIONS`. Your decision doc is accurate; the body is not. The fix lane is drafting a corrected body.

Retargeted to `dev` — that is my gap, not yours; #214 documents the branch and has not landed yet.

This is good work. The normalisation measurement in particular is the kind of thing that only shows up if you go looking for it.

PR #233, opened 2026-08-17
Nominate five Thrill Studio skills — two design, three craft

thrillmot:
Two design and three craft skills, incubated in a consumer repo and proven by use there. Per the nomination flow in `docs/integrating-with-agent-skills.md`.

| skill | body | what it is |
|---|---|---|
| `empirical-design-principles` | 7514 | Fitts, Hick, Miller, Jakob, von Restorff, Gestalt — the falsifiable half |
| `usability-heuristics` | 8180 | Nielsen, Norman, Shneiderman, progressive disclosure — the judgement half |
| `proving-an-absence` | 8149 | a probe that finds nothing may be the wrong probe |
| `guarding-a-regression` | 8168 | a guard you have not watched fail is not a guard |
| `retiring-a-superseded-decision` | 8168 | correcting a doc without hunting the old claim |

**A sixth is deliberately withheld.** `deploying-confidential-work` stays in our own repo: its evidence *is* our infrastructure, which is what makes it useful to us and unpublishable here.

## The disclosure check caught what two prior passes missed

This catalog is public, so that check mattered more than the quality one.

`proving-an-absence` carried a worked example that **reproduced a client's route registry almost verbatim** — `app/site-header.tsx`, `app/site-structure.ts`, searching for one of that project's own declared page types. Name and path greps did not find it, because nothing was *named*. Only reading it against the consumer repo's own docs did.

Neutralised to a generic registry. The example's point — a value that is generated, not written — is unchanged.

Four other genericising edits: a dead cross-reference to a client-specific skill rerouted to `consuming-a-design-system`, `applies_to` paths replaced with the house monorepo-safe globs, and `test-discipline` addressed as a sibling rather than as an external catalog.

## Reference links are absolute, deliberately

`empirical-design-principles` is **the catalog's first skill with a subdirectory** (`find skills -mindepth 2 -maxdepth 2 -type d` was empty before this). Three decisions you may want to overrule:

- **Subdirectories travel.** Verified against `anthropics/skills`: `npx skills add` copies `references/`, `scripts/`, `assets/` wholesale.
- **Links into them are absolute URLs, not relative.** Measured: a relative link 404s on skills.sh, which renders `SKILL.md` alone; the absolute form resolves there, on GitHub, and when installed. A link must work from every surface the skill is read on.
- **The 112 KB is reachable but not loaded** by a consumer that reads `SKILL.md` alone — that is #201, unresolved. Recorded in the placement-map note rather than papered over.

## Verified

```
validate_skills.py            OK: 54 skills validated cleanly
                              control: mutating a name → 2 errors, exit 1
disclosure scan               0 hits across all 8 files including references/
                              control: same probe on the consumer CLAUDE.md → 7 hits
commit SHAs                   0 real; 21 hex runs are all DOIs, PMIDs, ISBNs
                              control: a known short SHA resolves, these do not
bodies                        all ≤ 8192, nothing deleted to fit
prose retention               no undeclared removal
relative links                22, all resolve; control target resolves False
```

`docs/placement-map.json` gains five entries because `validate_skills.py` reconciles its keys 1:1 against `skills/` — five directories without five entries is five errors, not a nicety.

## One thing you should know before trusting a green check on this PR

**`check-links` does not walk `skills/`.** Mutation-tested:

```
broken link injected into skills/proving-an-absence/SKILL.md
  → "All markdown links resolve and no orphans found."   exit 0

broken link injected into docs/integrating-with-agent-skills.md
  → "Broken links (1): missing -> ./zzz-nope.md"          caught
```

So its clean output on this PR says nothing about these files. The Python resolver above is what actually checked them. Filed separately — it matters beyond this PR, because `check-links` is now a *required* status check.

## Left alone deliberately

No reciprocal backlinks added — each is a prose-retention-gated edit on a file outside this port's scope. And one is currently impossible: `test-discipline` has **1 byte** of headroom (8191/8192), so a backlink cannot be added without deleting prose.


thrillmot:
# Verdict: MERGE WITH FIX

Five fixes. Four are edits measured in lines and land in an hour. The fifth is a scoped citation audit — about one agent-day, and it is the only one with real duration. All five fit in three days with margin **if the audit starts today**.

I re-ran the load-bearing checks myself in an isolated archive of `1d0f848` rather than take them on report. Where I say "measured", I ran it; where a lens measured it and I did not, I say so.

---

## Required before merge

### 1. `empirical-design-principles` fires on Fitts and Jakob and its body contains neither word

Measured on the PR head, case-insensitive, frontmatter split with the same regex the repo's own validator uses:

```
term          desc  body        positive control (references/)
fitts            1     0        interaction-laws.md      Fitts = 32
jakob            1     0        convention-and-memory.md Jakob = 18
gestalt          1     2
hick             1     1
von restorff     1     1
zorkmid          0     0        ← negative control, discriminates
```

The description promises "Gestalt grouping and common region, Fitts's, Hick's, Miller's, Jakob's, von Restorff". Two of the seven are delivered only by files the body links to. An agent that triggers on "Fitts's Law" reads a body that never says the words.

**Fix:** one line each for Fitts and Jakob in the body, placed in the **front half**. Headroom, boundary-tested at the byte against the real validator — `+279` → `OK: 54 skills validated cleanly`, `+280` → `body is 8193 bytes, over the 8192-byte limit by 1`. So **279 bytes**, which is two short sentences. Front half matters because of fix 5.

This fix is also what makes the subdirectory ruling below defensible. Do not skip it.

### 2. The three `blob/main` reference URLs are 404 and will stay fragile

Live, `curl -L`, first and final status both recorded so no redirect could read as a false zero:

```
404  …/blob/main/…/references/gestalt.md
404  …/blob/main/…/references/interaction-laws.md
404  …/blob/main/…/references/convention-and-memory.md
200  …/blob/main/skills/apca-contrast/SKILL.md          ← positive control, same URL shape
200  …/blob/main/README.md                              ← positive control
404  …/blob/main/skills/ZZZ-CONTROL-DOES-NOT-EXIST/…    ← negative control
200  …/blob/1d0f848…/…/references/gestalt.md            ← the files do exist, on this branch
```

They resolve after merge to `main`. That is not the problem. The problem is that `SKILL.md:75` sends an installed reader to the network for files that are **already sitting next to it on disk** — `npx skills add` copies subdirectories wholesale, which the links lane proved with a fresh install of `anthropics/skills -s skill-creator` producing `.agents/skills/skill-creator/references/schemas.md`.

**Fix:** make the relative on-disk path the primary pointer (`references/interaction-laws.md`), keep the absolute URL as a fallback for a web reader, and SHA-pin it or drop it. A `blob/main` URL in a permanent public artifact is a promise that a path never moves.

### 3. Scope line 139's verification claim to what was actually verified

`skills/empirical-design-principles/SKILL.md:139` says: *"Attributions there were verified against the original publications, and several corrections to widely repeated citations are recorded inline."*

What supports it: **31 of 31 DOIs resolve to the correct work** via Crossref with a positive control (`10.1145/3313831.3376878` → correct title/venue/year) and a negative control (`10.9999/…` → NOTFOUND). Seven NN/g article URLs return 200 with titles matching the citations verbatim. Roughly 20 of ~200 citations were opened and all 20 were clean.

What does not: nobody opened Sampson & McGowan's 90.5%/82.5%, Yin et al.'s 14.8–24.4%, iComment's 60/33, Palmer 1992 pp. 438/439, Ke et al. 2023's four experiments, Peterson & Salvagio's 57%/89%, the Tesler/nomodes provenance, the Fitts 1954 equation forms, the MacKenzie 2018 ch. 17 quote, the ISO 9241-9:2000 → 9241-411:2012 renumbering, or the "none of whom is Wertheimer" attribution claim in the Sources section.

The DOI result bounds the error class to in-paper numbers rather than misattributed papers. That is a real and useful bound. It is not the sentence on line 139.

**Fix, in this order:** (a) audit the named list above — finite, about one agent-day, and it is the single largest unexamined surface in the PR at 1,262 lines; (b) regardless of what the audit finds, rewrite line 139 to state the scope of the verification rather than assert its completeness. A skill about evidence cannot be the artifact that overclaims its own. This is permanent and public and one client never gets a correction.

### 4. Log the decision

`check-decisions` passes here, vacuously. The gate counts non-docs lines and `skills/**` and `docs/**` are excluded, so this PR's countable total is 0 and the gate is satisfied without any decision file. 2,112 lines into a permanent public catalog — including a precedent-setting subdirectory ruling — with nothing in the log.

**Fix:** one `logmind log` entry. It is also the right home for the `references/` rule below.

### 5. Before merge, establish which byte cap the client repo runs under

`validate_skills.py:81-84`, this repo's own source: `8192` is clud-bug's `DEFAULT_MAX_SKILL_BYTES`, and *"Its shipped workflow templates override this DOWN to 4000, so a repo on an unmodified template truncates earlier — clud-bug#301."*

Four of these five bodies are over 8,100 bytes. Under a 4,000-byte consumer, roughly the back half of each is silently cut — for `empirical-design-principles` that is everything from the Cross-references block onward, including all three reference URLs. This is a handover question, not a code change, but it decides where fix 1's two lines go and whether fix 2's URLs are reachable at all. Read the client repo's actual `max_skill_bytes`. Do not infer it.

---

## Ruling on the subdirectory: **accepted**, and here is the rule that follows

The size gate's own error message says: *"Do NOT move prose into references/ — that consumer reads SKILL.md and nothing else, so the move deletes it. There is no exception list."* I fired it to read it verbatim. Left as-is next to a shipped `references/`, the catalog contradicts itself and skill #51 gets to pick which sentence to believe.

It does not contradict itself, because the message bans a **move**, for a stated **reason**. Prose relocated out of a body to duck the byte count is deleted from the reader who matters. That is not what happened here: these 1,262 lines were never instruction, and the body is at 7,912 of 8,192 — it did not need the room.

**The rule for skill #51.** A skill may ship `references/` when all four hold:

1. **Severability.** Delete `references/` and the SKILL.md is still a complete instruction — the agent can do the job without opening it, and *everything the description promises is honoured in the body*. This is the operative test.
2. **Kind.** What is in there is source material — citations, provenance, verbatim quotations, evidentiary-standing tables. Not procedure, not rules, not the skill's own judgements.
3. **Reachability.** The body points at it by relative on-disk path first, because the installer copies the directory and the reader has it locally. Absolute URLs are fallbacks, SHA-pinned or absent.
4. **Hand review, declared.** No gate in this repo reads a byte of it, so shipping the directory is a standing commitment to review it by hand on every change, re-syncs included.

And the inverse, which is the original rule intact: **`references/` is never where something goes because it did not fit.** Over 8,192, the answer is still cut, link to a neighbour, or split.

Note that `empirical-design-principles` **fails test 1 as submitted** — the description promises Fitts and Jakob, only `references/` delivers them. Fix 1 is what earns the ruling. Ship the fix or the precedent is unsound.

**Also required:** amend `validate_skills.py`'s message so "There is no exception list" reads as what it means — no exception for *moving instruction prose out*. Two lines, same commit. Leaving two contradictory rules standing is worse than either rule.

The gate blindness is real and I reproduced it. `SKILL_GLOB_RE = re.compile(r"^skills/[^/]+/SKILL\.md$")` at `check_prose_retention.py:123`. I cut `references/interaction-laws.md` from 464 lines to one and got `OK: 54 skills validated cleanly` and `All markdown links resolve and no orphans found`. Control: the lane cut a whole `## Nielsen` section out of a SKILL.md and the gate fired — *"Deleting prose is allowed. Doing it silently is not."* So the zero discriminates. 110 KB of the catalog's most citation-dense material can be gutted with every check green. Rule 4 above is the only thing standing between that and a silent loss; it belongs in the decision log, not in a reviewer's memory.

---

## Disclosure: clean, and here is exactly what "clean" covers

My own controlled sweep, 8 new files, 33-ish client and studio terms plus email, absolute paths, and the incubation branch name: **5 hits, none of them a leak.** Four are `thrillmade` inside this catalog's own public `github.com/thrillmade/agent-skills/…` URLs. The fifth is `/Home/` in `guarding-a-regression/SKILL.md:89` — the keyboard key, matched by my `/home/` path pattern. **Control, same regex, same code, 25 files from the client trees: 332 hits.** The probe finds client vocabulary when client vocabulary is there.

The lane's own passes, each with a firing control:

- Arlyn design vocabulary: 0 in target, **69** in `packages/catalyst/src/styles/typography.css` + a deck test.
- Pharma / market-access: 4 hits, all published literature — Redelmeier & Kahneman's "287 patients", Humphreys & Riddoch's "Balint's syndrome patient", and `prescri` matching "prescriptive". Control: 11 on client files.
- Email: 0. First control attempt returned 0 and was **discarded as broken**; replaced with two real ones, one of which surfaces the client's actual public address in the client tree.
- Ticket IDs, commit SHAs, absolute paths, branch names: 0. The incubated versions carried `e3441ec`, `fd4e94a`, `7b9738e`, `f0fd390`, `/vercel/path0/.next/package.json` and `feat/r2-deck`. All gone.
- Six-word shingles against the whole client corpus, 192 files / 248,543 shingles: **21**, every one enumerated — 10 are these skills' own description text quoted back by a client artifact (flow in the other direction), 2 are filenames in `skills-lock.json`. Control: the **incubated** versions of the same three skills against the same corpus, **88**, including `packages catalyst src styles typography css`. The method finds client text when it is present.
- Seven other client and studio trees, 388 files: 1 hit, `because it s still relevant to` — the Nygard ADR quotation and ordinary English.

**What "clean" does not cover, stated plainly:**

- **The 1,262 lines of `references/` had no disclosure pass at all.** I verified this: sha256 identical to the incubated copies for all three files, while all five SKILL.md files **differ** from their incubated versions — so the control shows the comparison discriminates and the scrub genuinely ran on the bodies and genuinely did not run on the references. They are clean, checked after the fact. That is an outcome, not a process, and nothing guards the next re-sync. Rule 4 above is the guard.
- The shingle corpus is working trees, not git history. Text deleted from the client repos before today would not appear. Incubation was 13–17 Aug, so the window is the relevant one, but it is a bound.
- `n=6` misses paraphrase under six consecutive words. Mitigated by a hand read of all 2,054 lines and five independent vocabulary greps, not eliminated.
- Binary assets were not inspected for embedded text. The eight files contain no images and no asset references, so there is no path from asset content to these files — a corpus gap, not an unchecked channel.
- **Nobody could confirm which two of the five ship to the client.** `arlyn-delivery/skills/` currently holds only a README (`c1c53e6` pulled them), and `skills-lock.json` lists none of these five. All five were therefore held to the same bar. Whoever owns the handover needs to name the two, because fix 5 depends on it.

---

## CI's green tells you nothing about the links in these files

`check-links` is a required status check and it does not read skill bodies. My own mutation pair, `logmind 1.2.0`, the same binary CI installs:

```
break ../guarding-a-regression/SKILL.md inside skills/proving-an-absence/SKILL.md
  → "All markdown links resolve and no orphans found."        BLIND

identical break in README.md
  → "Broken links (1): README.md: missing -> skills/ZZZ-…"    FIRES
```

Three lanes reproduced this independently, each with its own firing control. The workflow is a vendored `logmind` template (`logmind-template-version: v5`) whose own header scopes it to references *between docs*. Skill files are link targets, never sources. A separate lane also put a genuinely dead absolute URL in `README.md` — a file the check demonstrably reads — and got `rc=0`, so absolute URLs are unchecked everywhere. 293 relative sibling links across this catalog have never been validated by anything. #234 / #197.

**Verified by hand instead**, my checker, one code path, controls that fire:

- **24 relative links and 35 in-file anchors across all eight new files → 0 broken**, checking file existence *and* anchor slug. Control: injected `../ZZZ-NOPE/SKILL.md` → BROKEN, `#zzz-no-such-anchor` → BROKEN, `../guarding-a-regression/SKILL.md` → OK, same run. A separate lane widened this to 94 links including `README.md` and `docs/file-structure.md` → 0 broken.
- The three `blob/main` reference URLs → 404, with the controls in fix 2.
- 31/31 DOIs, controlled resolver. Seven NN/g URLs → 200, titles matching verbatim.
- `validate_skills.py` → `OK: 54 skills validated cleanly`, and the pass is real, not a file the gate never opened: injecting `": "` into a `description:` produced `frontmatter is not valid YAML … line 2, column 22`, exit 1.

**One trap for whoever re-checks after merge:** `skills.sh` soft-404s. `skills.sh/thrillmade/TOTALLY-BOGUS-CONTROL/SKILL.md` returns **HTTP 200** with the same title shape as a real page. A 200 there proves nothing; body size discriminates (42,639 vs 76,340 bytes). Post-merge, check the three reference URLs on `github.com` with `curl -sL -w '%{http_code}'` — 200 is the expected answer and anything else means the merge did not reach `main`.

---

## Filed separately — not blocking this PR

- **`check-decisions` counts vacuously.** `skills/**` and `docs/**` are excluded, so a pure-catalog PR of any size passes with zero decisions logged. Verified with two independent implementations of the exclusion filter: this range → 0 countable, `e33ad45` (previous main merge) → 6,632. The 0 is a property of the PR, not a broken probe. Fix the gate; fix 4 covers this PR.
- **`AGENTS.md:35` says the suite is "112 tests" and `:39` says both commands "run in about a second".** I collected **570** on this tree in 0.23s; the full run is minutes, not a second. Not caused by this PR — it is the first thing the next agent reads, and it is wrong by 5x.
- **`README.md:34` sends public readers of a public catalog into a private tracker.** `github.com/thrillmade/protocol/issues/39` anonymously → **404**; control `github.com/thrillmade/agent-skills/issues/233` → **200**; the org repo itself → 404, so `protocol` is private wholesale. Every stranger who follows that link hits a wall. Small, and worth doing before the catalog gets more traffic.

---

## Credit

Worth saying because it is unusual and it is measurable.

**The lane found and reported a leak that two review passes had missed, and it withdrew two skills as client-specific rather than scrubbing them.** The 88 → 21 shingle drop between the incubated and submitted versions is that work, and the fact that 10 of the surviving 21 turn out to be the client artifact quoting *these skills* rather than the reverse is the shape you want. Four commit SHAs, a `/vercel/path0` build path, and a working branch name were carried by the incubated copies and are gone from these. Deciding that two of seven candidates were not extractable is the judgement that is hard to make and easy to skip.

Two lanes also caught their own broken probes mid-run and said so — an email control that returned 0 because it was pointed at guessed-at files, and a `check_prose_retention` control that "passed" only because it ran with no diff to see. Either one, unreported, would have put a false clean in this review. That is the discipline this catalog is supposed to teach, applied to the catalog itself.

---

## Summary

| | |
|---|---|
| **Verdict** | MERGE WITH FIX |
| **Blocking** | Fitts/Jakob into the body (279 bytes headroom, front half) · relative-path pointers, SHA-pin or drop the `blob/main` URLs · scope line 139's verification claim + audit the named citations · log the decision + the `references/` rule · establish the client's real byte cap |
| **Subdirectory** | Accepted, under the four-part rule above; `validate_skills.py`'s message amended in the same commit |
| **Disclosure** | Clean on eight controlled probes; the 1,262 reference lines were never scrubbed and are clean by luck, not process |
| **CI** | Green is silent on every link in these five skills; 24 relative links, 35 anchors and 31 DOIs checked by hand instead, 0 broken, 3 URLs 404 |
| **Three days** | Yes, if the citation audit is dispatched today |

thrillmot:
## Three asks before we finish this — two are yours alone

The verdict and the five fixes are in the comment above. These are the questions the panel could not
answer from the outside. I am measuring what I can independently in parallel, so where you answer and
I measure, we get a cross-check rather than a substitute — please answer from what you know rather
than checking what I would find.

### 1. What `max_skill_bytes` does the client repo actually run?

This is the one with real consequences. Four of the five bodies are over 8,100 bytes. `validate_skills.py:81-84`
says clud-bug's shipped workflow templates override the cap **down to 4000**:

> Its shipped workflow templates override this DOWN to 4000, so a repo on an unmodified template
> truncates earlier — clud-bug#301.

If the consumer is on 4000, roughly the back half of four of these skills is **silently cut** — for
`empirical-design-principles` that is everything from the Cross-references block onward, **including
all three `references/` URLs**. The skill would then point at its own source material with a line the
reader never receives.

**Please read the number and tell us the file it came from**, rather than the value you expect. If it
is unset, say unset and name the file you checked. This decides where the two required Fitts/Jakob
lines go, and whether fix 2's reachability change is even load-bearing.

### 2. Which two of the five actually ship to the client?

The panel held all five to the same bar because it could not determine this — `arlyn-delivery/skills/`
currently holds only a README (`c1c53e6` pulled them), and `skills-lock.json` lists none of the five.

You said two of these ship to a repo handed over on the 20th and never refreshed. **Which two?** The
permanence argument applies with full force to those and only partly to the rest, and fix 3's citation
audit should be sequenced to cover them first if the three days get tight.

### 3. An ELI5 of each of the five, in your words

Our CEO is deciding whether these go into a public catalog under the studio's name, and he has not
read them. For each of the five: **what is it, in two or three sentences a smart non-specialist reads
once and gets** — no catalog vocabulary, no "L0"/"lens"/"dispatcher". Concretely, prefer *the thing it
prevents* over *the category it belongs to*.

And one honest line each: **general craft, or house habit wearing a general name?** You already
withdrew two as client-specific, so you have clearly been applying that test — we would rather have
your read than infer it. A skill that is genuinely ours-only is a fine answer and costs nothing to
say; the expensive version is finding out after it is public and permanent.

---

Nothing here blocks you starting the five fixes. Fix 1 (the Fitts/Jakob lines, 279 bytes of headroom,
boundary-tested) and fix 3a (the citation audit, ~one agent-day) are the two with duration — those are
worth starting today regardless of how these three questions come back.

thrillmot:
## Correcting my own fix 5 — I measured it instead of waiting, and the 4000 threat is not real

I asked you to read the client repo's `max_skill_bytes`. I then measured it across every repo on this
machine, and the answer removes fix 5 rather than answering it. **Two of my three questions above are
now withdrawn.**

### The effective cap is 8192, and 4000 reaches nothing that consumes this catalog

| consumer | cap | where |
|---|---|---|
| **clud-bug App** — the reviewer that actually runs here | **8192**, unset everywhere | default `clud-bug-app/lib/env.ts:244-250`; enforced `clud-bug/src/core/prompt-builder.ts:55`, sliced `:302-303` |
| producer gate | **8192** on body | `validate_skills.py:85`, applied `:215-216` |
| **arlyn-working** — the real consumer | **none** | no `.github`, no clud-bug; Claude Code's own loader. An installed 9047-byte `visual-polish` is intact, tail and all |
| clud-bug-test (throwaway smoke repo, consumes none of the catalog) | 4000, **advisory only** | `.github/workflows/clud-bug-review.yml:227` — interpolated into *prompt prose* at `:249-252`, `:297`, `:354` telling the model to `head -c` itself. **Nothing in code enforces it** |

`vercel env ls production` returns 27 vars; `MAX_SKILL_BYTES` is not among them. And
**`.clud-bug.json` is not a control surface for this** — nothing in the codebase reads a cap out of the
manifest, so setting one there is inert.

So `validate_skills.py:83-84` calling 4000 an "override" is misleading: it is a prompt instruction, not
a mechanism, and **no catalog consumer is on that template.** Filing that separately.

### The real finding is the opposite of the one I raised: headroom, not truncation

| skill | body | headroom |
|---|---|---|
| empirical-design-principles | 7912 | 280 |
| proving-an-absence | 8148 | 44 |
| guarding-a-regression | 8167 | 25 |
| retiring-a-superseded-decision | 8167 | 25 |
| **usability-heuristics** | 8179 | **13** |

The gate and the consumer are **the same number with zero buffer**. Nothing gets truncated today; the
exposure is that a typo fix flips a skill from passing to over-limit. Note fix 1 needs 279 of
`empirical-design-principles`' 280 — that is not a coincidence I would want to rely on twice.

*(Two independent agents measured these one byte apart — 7912/7913 etc., consistently off by one,
likely a trailing-newline convention. Immaterial at 280; **not** immaterial at 13. Worth someone
settling which convention the gate actually applies before editing `usability-heuristics`.)*

### Question 2 also resolved, and it contradicts the premise

**Zero of the five reach `arlyn-delivery`, and none ever has.** `git log --all -S'<slug>'` returns 0
commits for each of the five; control `-S'apca-contrast'` → 3 commits, so the probe finds a skill when
one is there. The only `skills-lock.json` that ever existed there pinned 17 skills, none of these five.
At HEAD it is a README-only stub, and `arlyn-delivery/CLAUDE.md:27-38` says the design-system skills are
**held back deliberately**, pending the token generator and per-skill catalog versioning.

**This matters because the whole review — mine included — ran on "two of these ship to a client on the
20th and are never correctable."** That framing set the severity bar on every lens. On the evidence it
does not currently hold.

I am not calling you wrong: you know the intent and I only see the tree. But please reconcile it —
either name the mechanism by which these reach the client before the 20th, or confirm the handover
ships without them. **If it is the latter, the deadline pressure comes off fix 3's citation audit**,
which is the one item that could not be done well in the time.

Worth knowing: the per-skill versioning `arlyn-delivery/CLAUDE.md` is waiting on **is built** — digest
stamped in frontmatter, a checker a subscriber runs in their own repo, 824 tests. Not yet reviewed or
merged, but it is no longer the blocker that note assumes.

### Still open, and now the only one I need from you

**Question 3 — the ELI5 of each of the five in your words**, and the honest general-craft-vs-house-habit
line. I have my own read and would rather compare than substitute.

thrillmot:
## I am taking the five fixes — do not push to this branch until I say it is clear

**Collision warning first, since this is your branch:** I have a lane working on
`feat/nominate-thrill-studio-skills` right now. **Please hold off pushing to it.** I will comment when
it is clear. If you have local work on it, say so now and I will take your commits instead of writing
over them.

## Two changes to the PR itself

**1. Retargeted to `dev`.** It was opened against `main`, which is the default `AGENTS.md` explicitly
warns about — *"The default base is `main`, so set it by hand."* Two practical effects: it now follows
the convention, and it moves inside my merge authority, so this no longer waits on the CEO. Lossless —
the branch's merge-base with `dev` and with `main` is the same commit (`e33ad45`).

*(The first attempt failed with a Projects-classic GraphQL deprecation error and I nearly read that as
cosmetic. It was not — the base was unchanged. Redone through the REST API.)*

**2. Fix 5 is withdrawn, because I measured it instead of asking you.** I asked what
`max_skill_bytes` the client repo runs. The answer removes the question:

| consumer | cap |
|---|---|
| clud-bug App — the reviewer that actually runs here | **8192**, unset everywhere, code default |
| producer gate | **8192** on body |
| `arlyn-working` — the real consumer | **none**; no clud-bug at all, Claude Code's own loader |
| clud-bug-test (throwaway, consumes none of the catalog) | 4000, **advisory only** — interpolated into prompt prose telling the model to `head -c` itself; nothing in code enforces it |

`vercel env ls production` returns 27 vars and `MAX_SKILL_BYTES` is not among them. **The 4000 reaches
nothing that consumes this catalog**, so the truncation concern I passed to you does not apply. The
real exposure is the opposite one — headroom. `usability-heuristics` has **13 bytes**.

## What I confirmed before starting

Your description-vs-body gap reproduces on my own instruments:

```
skills/empirical-design-principles/SKILL.md   body 7912 bytes, headroom 280
'fitts'  frontmatter=1  body=0
'jakob'  frontmatter=1  body=0
CONTROL 'gestalt' frontmatter=1 body=2     <- the probe finds a term that IS in the body
```

That is the one fix that earns the subdirectory ruling; a skill that fails its own severability test
cannot be the precedent for shipping `references/`.

## Still yours, and it is the one question I cannot answer from the tree

**Which of the five actually reach the client, and by what mechanism?** Measured: zero of the five have
ever reached `arlyn-delivery` (`git log --all -S'<slug>'` returns 0 for each; control `-S'apca-contrast'`
returns 3), its `skills/` is a README-only stub, and its `CLAUDE.md:27-38` says the design-system skills
are held back deliberately pending the token generator and per-skill versioning.

The CEO has since confirmed the *intent* is to ship design principles to the client, so I am treating
the deadline as real. But **the mechanism does not exist in the tree yet**, and if the handover ships
without them the time pressure comes off the citation audit — which is the item that most needs the
time. Your call, and it changes what I prioritise in the next few hours.

thrillmot:
## Verdict: MERGE WITH FIX

Four of five fix-claims held through the rebase, verified against the real diff, the real tree, and the real generator code -- not against the PR's own report of itself. One did not: a note added by this very PR in `docs/placement-map.json` asserts something about `empirical-design-principles/SKILL.md` that the file no longer does. That's a one-line fix, not a design problem, hence MERGE WITH FIX rather than DO_NOT_MERGE.

Reviewed from an isolated archive of `origin/feat/nominate-thrill-studio-skills` at `33dbdbf` (base `origin/dev` at `b6d67cb`), diffed as `origin/dev...origin/feat/nominate-thrill-studio-skills`. Nothing below is claimed about a file outside that diff.

### 1. Severability fix -- HELD

`empirical-design-principles`'s description promises seven principles: Gestalt/common region, Fitts's, Hick's, Miller's, Jakob's, von Restorff, aesthetic-usability. Grepping the BODY only (everything after the frontmatter close), with controls:

```
proximity (known-present control): 2
xyzzynotreal (known-absent control): 0
common region: 2   Gestalt: 1   Fitts: 1   Hick: 1   Miller: 1   Jakob: 1
von Restorff/Restorff: 1   aesthetic-usability: 1
```

All seven now clear at least one substantive body mention. Fitts's and Jakob's -- previously frontmatter=1/body=0 per the prior panel round -- now live in "## Two that hold, inside a stated scope" with their actual predicate (Fitts: "rises with distance, falls with target size, logarithmically," scoped to known-location targets; Jakob: scoped to control placement/behaviour, visual half flagged as Nielsen's opinion, not measured support). Fixed and held.

### 2. Reference reachability -- links resolve, but a fresh metadata note contradicts them (BLOCKING)

Extracted every `](...)` link from the 5 nominated skills' `.md` files (29 total across `SKILL.md` + the three `references/*.md` files) and resolved each against the archived tree. Control-tested first: a deliberately broken link (`references/does-not-exist.md`) was correctly flagged `FAIL`, confirming the checker isn't a rubber stamp. Result on the real files: **29/29 OK**, zero `FAIL`. No `blob/main` URL remains anywhere in the 5 skills except one line of prose in `empirical-design-principles/SKILL.md` explaining *why* none is given ("No `blob/main` URL is given: that promises a path that never moves").

That's the fix working as designed -- for on-disk / `npx skills add` installs, which is what the SKILL.md text itself argues for ("`npx skills add` copies `references/` wholesale").

But `docs/placement-map.json` gained a note in this exact PR (diff shows it added fresh, not carried over) that says the opposite is true of the same file:

> "so its SKILL.md links them by ABSOLUTE URL - skills.sh renders SKILL.md alone and 404s a relative reference link"

This is stated as present-tense fact about a file that, in the same commit, uses relative links and says so in its own prose. `skills.sh` is a real, active distribution channel this repo serves (README: "These skills run inside any agent that loads skills.sh"; it's the default install path documented throughout `docs/`). Either:
- the note is stale and should say what the file actually does now, or
- the underlying claim is real and the relative-link fix creates a genuine 404 on skills.sh that nobody has verified either way.

I could not check which, because I have no access to how skills.sh renders a skill page. What I can say is that the PR ships both claims at once, in the same commit, about the same file -- exactly the "hand-kept second copy" AGENTS.md warns reads as true until it quietly isn't. Recommend: update the note to match the shipped form, and if the skills.sh risk is real, record it as an open question rather than a settled "so it does X."

### 3. Scoped verification claim -- no longer overclaims

Current text (was previously "verified against the original publications," per the task brief):

> "Verification is partial and the scope is the claim: about 20 were opened and read against the original... and the rest are DOI- or URL-resolved only. A resolving DOI shows a work exists, not that it says what the citing line claims."

This is honestly scoped, not a blanket claim. Spot-checked what's checkable:
- **"78 source entries"** is exact -- counted bullets under each reference file's own `## Sources` heading directly: `convention-and-memory.md` 26 + `interaction-laws.md` 28 + `gestalt.md` 24 = **78**.
- Explicit `[read in full]` / `[read in full and text-searched]` tags: 4, all in `interaction-laws.md` (Fitts 1954, Liu et al. 2020, Miller 1956, Doherty & Thadani 1982).
- The claim's "plus the ones the inline corrections turn on" clause is independently supported: `gestalt.md` carries page-numbered direct quotes (Palmer p.446, p.439; Wagemans et al. p.1190, p.1206) and an explicit *non*-verification flag ("could not be verified against the text," on a Koffka page cite) -- the kind of granular hedge you only get from actually opening the source, in both directions.
- Could not reproduce "about 20" to the exact digit (would require reading all 78 entries against their primaries); the pattern of evidence is consistent with, not contradicting, that figure. Reported as not-fully-checked rather than assumed true.

### 4. Rebase artifacts -- clean

- `docs/placement-map.json`: all 5 skills carry `family`/`owns`, and all 5 `family` values resolve against the declared `families` list (`design-primitives` x2, `review-discipline` x2, `catalog-and-tooling` x1).
- Generated directory `skills/finding-a-catalog-skill/SKILL.md` lists all 5 under their correct sections (design primitives, review-discipline, catalog-and-tooling).
- `python3 .github/scripts/gen_skill_directory.py --check` → `OK: ... is current` (exit 0) on the real tree. Control-tested: mutated a copy of the generated body and reran -- correctly refused (`the committed directory body differs...`, exit 1) before trusting the clean pass.

### 5. The "not here" probe narrowing -- re-derived independently, narrowing was correct and necessary (not just stylistic)

Imported the PR's own `_matching_files()` from `gen_skill_directory.py` and ran it directly against the archived `skills/` tree (not a re-implementation):

```
'forcing function'  -> 1 match: ['empirical-design-principles']
'natural mapping'   -> 1 match: ['empirical-design-principles']
'constraints'       -> 1 match: ['empirical-design-principles']
'gulf of execution' -> 0 matches
'gulf of evaluation'-> 0 matches
'APCA' (control)    -> 8 matches (matches the generated file's own claim)
'xyzzynotreal' (neg. control) -> 0 matches
```

This confirms the premise exactly: `empirical-design-principles`'s own frontmatter description contains the phrase "forcing functions" (as a *disclaimer* -- "are in neither skill and are covered nowhere yet" -- not a claim of coverage), but the matcher only sees a substring hit, not intent. Left as `forcing function`, the probe would now report `filled` and `not_here()` would `SystemExit` -- refuse to generate the catalog directory at all, per the code's own logic (`if filled: raise SystemExit(...)`). This isn't a style call; the old probe would hard-break `gen_skill_directory.py` today. `natural mapping` and `constraints` would fail exactly the same way, so neither was a viable alternative target. `gulf of execution` and `gulf of evaluation` both measure genuinely 0 -- the gap the gloss text ("Norman's two gulfs -- execution and evaluation") describes is real on both halves, not just the one actually re-probed.

**Verdict on the narrowing: correct, and better than the alternatives.** Deleting the entry would discard a still-true, machine-verified claim about a real catalog gap (nothing here covers the gulfs). Leaving it unchanged breaks the generator, verified above. Narrowing was the only option that keeps the claim true and the tooling running.

One non-blocking gap in the self-check worth filing: the code only programmatically re-probes `gulf of execution`; the gloss's second half ("...and evaluation") is asserted in prose but not independently re-measured by `not_here()`. Both are 0 today (confirmed above), so nothing is wrong now, but a future skill that covers only the evaluation gulf would not trip the refusal that protects the execution half. Not a reason to block this PR -- the mechanism this PR relies on works today, control-tested -- but worth a follow-up so "both gulfs" stays a checked claim rather than a partially-checked one.

### 6. Gates, suite, links by hand

- `python3 .github/scripts/validate_skills.py` → `OK: 56 skills validated cleanly.` (exit 0)
- Full suite, isolated archive, dedicated venv: `pytest tests/ -q` → **704 passed in 260.30s**, exit 0. Command and result reproduced twice (once via Monitor-watched background run, once via a second concurrent run) with matching pass counts.
- README table: both new rows for `empirical-design-principles` and `usability-heuristics`, and the three added in the prior round (`proving-an-absence`, `guarding-a-regression`, `retiring-a-superseded-decision`), link correctly; all six link targets (5 `SKILL.md` + 1 `references/` dir) exist on disk.
- `docs/file-structure.md`: all 5 new entries are in correct alphabetical position within the `skills/` tree listing.
- `docs/timeline.md`: decision count incremented 22→23, newest entry points at `docs/decisions-branches/feat__nominate-thrill-studio-skills.md`, which exists and its content (measured `fitts`/`jakob` counts, the four `references/` tests, the Sources-claim history) matches what's independently verifiable in the diff.
- `check-links` (CI) is known, per AGENTS.md's own filed note (#234), not to read `skills/` at all -- so its status says nothing about the 29 links checked by hand above regardless of what it reports on this PR.

### Cross-reference sanity (not in the checklist, checked anyway)

The five non-nominated-skill edits (`designing-a-design-system`, `orchestrating-elite-agent-qa`, `reviewing-design-work`, `semver-design-tokens`, `visual-polish`) all add correctly-targeted links to the new skills with content that matches what those skills actually say (e.g. `visual-polish` pointing the spacing-ratio question at `empirical-design-principles` rather than re-deriving it). No orphaned or mismatched cross-references found.

---

**Net: fix this PR's one shipped-fresh factual contradiction in `docs/placement-map.json` (item 2), and it's clean to merge.** Everything else -- the severability fix, the 29 on-disk links, the honestly-scoped Sources claim, the placement-map/family data, the directory regeneration, and the probe-narrowing decision -- held through the rebase and is independently reproducible with the commands shown above, not merely reported.

Issue #237, opened 2026-08-17
Before CATALOG_HASH_ALGO is ever set: compute_catalog_hash hashes one file, and #233 ends the one-file-per-skill invariant


thrillmot:
`compute_catalog_hash()` hashes **one file**. It is correct today only because every skill directory
holds exactly one file — a property PR #233 ends.

Surfaced by the per-skill-versioning lane, which independently confirmed the live algorithm
(`sha256(relpath ‖ raw)`, files sorted by relpath) at **40/42** across four subscriber locks and then
*declined* to register it, because registering it would have been silently wrong the day a skill gains
a `references/` file.

### Nothing is broken today — this is a precondition, not a defect

`.github/scripts/census_counters.py:193`

```python
CATALOG_HASH_ALGO: str | None = None
```

With `None`, `compute_catalog_hash` returns `None` and every caller must treat the subscription as
**indeterminate**, never as drift. The module is emphatic about this and it is the right design:

> we NEVER report drift we cannot prove

So this issue asks for nothing to change now. It records **what must be true before
`CATALOG_HASH_ALGO` is ever set.**

### The failure it would introduce, and its direction

`census_counters.py:196-217` — the signature takes a single path, and the body opens exactly one file:

```python
def compute_catalog_hash(skill_path: str, algo: str | None = None) -> str | None:
    """Recompute the content hash of a CURRENT catalog SKILL.md …"""
    …
    with open(os.path.join(os.getcwd(), norm), "rb") as fh:
        raw = fh.read()
```

For a multi-file skill it would hash `SKILL.md` and ignore everything beside it. A subscriber whose
`references/gestalt.md` had drifted would be reported **`current`**.

**That is the false-green direction** — the one that matters. A false "stale" costs someone a
needless refresh; a false "current" means a subscriber is told they are up to date while they are not,
which is the whole failure this mechanism exists to catch.

### Measured state

```
skills/ subdirectories today            : 0
non-SKILL.md entries under skills/*/    : 0
files PR #233 adds under a skill dir    : 3   (empirical-design-principles/references/*.md)
```

So the invariant holds at HEAD and #233 is the first change to end it.

### This does not block #233

The subdirectory question is being ruled on separately and this is an input to it, not a veto:
`CATALOG_HASH_ALGO` is `None`, so merging #233 changes no reported verdict. What it changes is that
**setting `CATALOG_HASH_ALGO` is no longer a one-line change** — it now also requires
[...]

Issue #221, opened 2026-08-15
Upstream: logmind's file-structure generator reads the working dir — blocked on thrillmade/logmind#355 and #285, both open


thrillmot:
`docs/file-structure.md` is generated from a walk of the **working directory** rather than from git's view of tracked files, so machine-local scratch files land in a tracked, shared doc.

## Measured

A local logmind run staged this into `docs/file-structure.md`:

```diff
 ├── .claude
 │   ├── commands
 │   ├── skills
+│   ├── scheduled_tasks.lock
 │   └── settings.json
```

`.claude/scheduled_tasks.lock` is excluded from version control:

```bash
$ git check-ignore -v .claude/scheduled_tasks.lock
.git/info/exclude:8:**/.claude/scheduled_tasks.lock	.claude/scheduled_tasks.lock
```

## The precise boundary

The generator **does** honour `.gitignore` and **does not** honour `.git/info/exclude`:

```bash
# no .gitignore entry appears in the generated doc
$ for p in $(grep -v '^#' .gitignore | grep -v '^$' | tr -d '/'); do
    grep -q -- "$p" docs/file-structure.md && echo "PRESENT: $p"; done
(no output)

# control — a name that IS in the doc is found by the same grep
$ grep -c "settings.json" docs/file-structure.md
1
```

That is the worst half to miss. `.git/info/exclude` is **per-clone and untracked**, so it holds exactly the files that differ between machines — daemon state, locks, worktree scratch, agent registries. Those are the ones that should never reach a shared doc, and they are the ones that do.

## Why it matters beyond noise

The diff was sitting **staged and uncommitted** in the working tree. Staged-but-uncommitted derived-doc churn in a shared tree is how another lane's work rode into #212. A generator that emits per-machine content into a tracked file manufactures that state on every run.

## Owner and timing

This is a logmind generator behaviour, not an agent-skills one. **Raising it here rather than in logmind on purpose — logmind is heads-down on 2.0.0 and this is not worth interrupting that.** Reassign or move it after 2.0 ships.

Suggested fix when it is picked up: derive the tree from `git ls-files` (which respects every exclude mechanism by construction) rather than from a filesystem walk plus a hand-rolled `.gitignore` reader.

thrillmot:
**Third occurrence today, and the trigger is narrower and worse than I first described it.**

[...]

PR #238, opened 2026-08-18
Per-skill content versioning: a stamp a subscriber can recompute, and an index to look it up in

thrillmot:
Per-skill content versioning: a stamp a subscriber can recompute, and an index they can look it up in.

Answers the question a subscriber cannot currently answer — **is my copy of this skill current?** — including when the copy arrived by hand-copying or download rather than through `npx skills add`, because the digest is recomputed from bytes rather than tracked by the installer.

## The footer was refused, with the measurement

The obvious shape is a footer line in the body. It does not fit, at any size:

| footer | skills it breaks | body prose that must be deleted |
|---|---|---|
| 20 B | 3 | 38 |
| 35 B | 4 | 84 |
| 48 B | 4 | 136 |
| 75 B | 8 | 322 |

Tightest headroom is `test-discipline` at **1 byte**. Every footer is paid for by deleting body prose from the four tightest skills, each needing a `docs/prose-removals.md` row whose only honest reason is "to fit a footer" — and that file exists *because* "three skills lost real content to a link-conversion sweep that needed to fit under the 8192-byte size gate."

Two further measurements against it: the size cap **does not charge frontmatter** (`validate_skills.py` measures `body = content[m.end():]`), verified — stamping added 119 bytes to every file and changed **0 body bytes**, min headroom still 1. And a footer is the *last* bytes of the body, where a consuming reviewer truncates first.

## What ships instead

```yaml
version: "56be4995fc2f"  # is your copy current? github.com/thrillmade/agent-skills/blob/main/docs/skill-versions.json
```

**One owner per fact.** The digest is not repeated: the frontmatter line owns identity *and* the route, is generated by `stamp_versions.py`, and the validator requires it to be byte-identical to `skill_version.version_line(digest)`. Drift is unrepresentable rather than discouraged. The line is elided before hashing, so it is a fixed point — adding it changed no digest.

**Prose removed: 0 words. Ledger rows added: 0.**

## Live subscriber runs

```
arlyn-delivery @201e3c2~1   17 checked:  1 current, 16 STALE            exit 1
arlyn-delivery @c1c53e6~1   17 checked: 17 current                      exit 0
arlyn-working  @922eecb~1   30 checked:  2 current, 21 STALE, 7 local   exit 1
arlyn-working  @ccf8678     29 checked: 23 current, 6 local-skill       exit 0
```

`922eecb` is titled *"Refresh the 23 subscribed skills"* — the checker **stayed red because that commit shipped only symlinks**, going green at `ccf8678`. **0 false accusations**, audited across 11 repos / 86 installed files by re-deriving every verdict from bytes by an independent route.

## Verification

**824 tests** (570 baseline + 254 added). 24 guards mutation-proved. **Three failed on first attempt and all three were real** — including one where a guard was flipped to `pass` and 824 tests stayed green because nothing exercised `restamp`'s self-checks. One assertion was **deleted** because no mutation could reach it.

## The rebase proved the currency gate on a real mutation

Cut from `e33ad45`; `dev` has since taken #232, which changed one word in `frontend-a11y`. Because the stamp is a content digest, that made exactly one stamp stale, and the gate caught it unprompted:

```
::error file=skills/frontend-a11y/SKILL.md::`version` claims 766941312459 but this file's
content digest is 56be4995fc2f. The stamp is stale, so every subscriber comparing against
it reads a wrong identity. Run `python3 .github/scripts/stamp_versions.py --write`.
```

`stamp_versions.py --write` restamped **1 of 49** — precise, not a blanket rewrite. Generator verified **idempotent**: re-running changes nothing, tree stays clean.

## Reviewer notes

- **`docs/file-structure.md` arrived with its root node renamed to the worktree directory** (#221, fires on every worktree rebase). Repaired by hand; the legitimate additions in the same diff were kept.
- **Declined:** retiring `CATALOG_HASH_ALGO`. The algorithm was independently confirmed at 40/42 across four locks, but `compute_catalog_hash()` takes a *single file path* and is correct only because 0 of 49 skill dirs hold anything but `SKILL.md`. Filed as **#237** — PR #233 is the change that ends that invariant.
- **Ratification** of the `version:` key against protocol §2.1 needs filing on `thrillmade/protocol`. §2.1 sanctions unrecognised keys, so adding is sanctioned and *requiring* is not.
- `docs/skill-versions.json` 404s on `main` until this promotes. URL *form* control-tested.
- Every skill edit now needs `gen_skill_versions.py --write`, or the currency gate fails the PR. The error message says so.

**Needs an independent adversarial panel before merge** — the builder does not review its own work.


thrillmot:
## Yes — take the departed-repo lens, and it is the one my panel does not cover

An independent four-lens panel is running now: digest soundness, whether the subscriber checker lies,
whether the 24 guards actually guard, and the blast radius of touching 49 files. **None of them is the
departed-repo case.** You own that scenario and I do not, so please take it.

Here is what I have already established, so you can attack rather than rediscover.

**The pull path does work for a repo outside the org.** Measured:

```
curl -fsSLO https://raw.githubusercontent.com/thrillmade/agent-skills/main/.github/scripts/skills_current.py
python3 skills_current.py    # from your repo root
```

Stdlib only — `urllib`, `hashlib`, `json`. No auth, no dependency install, no org membership, no App.
`agent-skills` is public, so a private client repo outside the org can reach both the script and the
index. **That is the only half of the loop that survives a handover**, and I think it is the strongest
thing in this PR.

**The thing I would attack first, in your position: `skills_current.py` hardcodes `main`.**

```python
"https://raw.githubusercontent.com/thrillmade/agent-skills/main/"
```

For a departed repo that means the update channel *is* whatever sits on `main`, forever, with no way
to pin, and no signal if the URL shape ever changes. Worth asking what happens to a client running
this in six months against a repo that has been renamed, made private, or restructured — every one of
those is a silent failure that reports as an inability to fetch rather than as staleness. **A checker
that cannot distinguish "you are current" from "I could not find out" is the failure this repo keeps
rediscovering.** I have not tested that path; please do.

Other angles worth your attention on that case specifically:

- **Symlinks.** A real commit in `arlyn-working` titled *"Refresh the 23 subscribed skills"* stayed
  red because it shipped only symlinks, going green at the follow-up. Verify the checker treats a
  symlinked skill correctly for a handed-over tree, where the symlink target may not travel.
- **`local-skill` classification.** A departed repo will accumulate its own skills. Does one get
  misreported, and in which direction? False-`STALE` on a client's own work is noise they cannot
  action; false-`current` on a catalog skill is worse.
- **What it costs them to run.** Is there a story for a repo with no Python 3? That is not
  hypothetical for a client handoff.

**One correction to the framing in your prompt**, since it affects how hard you should push on the
20th. I measured `arlyn-delivery`: **zero of #233's five skills reach it, and none ever has.**
`git log --all -S'<slug>'` returns 0 commits for each; control `-S'apca-contrast'` returns 3. Its
`skills/` is a README-only stub and `CLAUDE.md:27-38` says the design-system skills are held back
deliberately, pending the token generator and per-skill catalog versioning.

I am not calling you wrong — you know the intent and I only see the tree. But please reconcile it on
#233: either name the mechanism by which those reach the client before the 20th, or confirm the
handover ships without them. **If it is the latter, the deadline pressure comes off #233's citation
audit**, which is the one item that could not be done well in the time remaining.

Worth knowing either way: the per-skill versioning your `CLAUDE.md` says it is waiting for **is this
PR**. Not merged, but no longer unbuilt.

thrillmot:
## Verdict: MERGE WITH FIX

Five items below must land before this enters `dev`. None is in the design — the design holds. Four are one-to-five-line changes, one is a missing test. I could not manufacture a false `current` on real data, and I tried hard. What I could manufacture is a **silent all-clear** — exit 0 on a copy that is genuinely behind — by three routes, one of which sits in the half of the index this PR itself declares ungated.

---

### What I re-derived vs. what I took on report

Isolated `git archive` of `origin/feat/per-skill-versioning` @ `0ec62965`. Nothing written under any repo in `/Users/ludlow/code/thrillmade/`; the sibling repos were read only.

**Re-derived, mine:**

| Claim | Result |
|---|---|
| 824 tests | `824 passed in 259.83s`, `PYTEST_EXIT=0` (echoed on the next line, not through a pipe). `--collect-only` → 824. |
| 49 skills / 49 index rows | 49 live, 49 published, 0 asymmetry either way. All 49 `current` equal `skill_version.digest(read_bytes())`. All 49 stamps equal their own file's digest. |
| "stamping added 119 bytes, changed 0 body bytes" | Per-file byte-delta set across all 49 is exactly `{119}`. Deleting the `version:` line from all 49 head files reproduces `origin/dev` **byte-for-byte, 0 exceptions**. Control: +1 byte on one file breaks the round trip. |
| "the size cap does not charge frontmatter" | `validate_skills.py:273` `body = content[m.end():]`, `:283` measures that. Max body 8191 (`test-discipline`), cap 8192 → **1 byte of headroom**. |
| "0 false accusations" | Independently, across 12 sibling repos: 135 installed skill dirs, 92 distinct copies examined, **0 silently dropped** (the 135→92 gap is entirely correct symlink dedup). Every `current` is byte-truth. |
| Diffstat | 64 files, +3378/−3. PR body says 62/+3369. Branch moved; immaterial. |
| CI | All seven real checks green. `clud-bug-review` = `neutral`, "clud-bug review unavailable" — the documented outage, so this panel *is* the independent review, not a second opinion on one. |

**Not re-derivable — taken on report or refused:**

- **"24 mutation-proved guards."** There is no committed mutation harness for these modules. `tests/test_prose_retention_mutations.py` is pre-existing on `dev` and touches none of them. The count describes a terminal session, not an artifact. I ran my own mutations instead (below); the guards I probed do have teeth, but nobody can re-run the 24.
- **"Three guards failed mutation on first attempt and all three were real."** Historical. No artifact. Uncheckable, and I am not calling it either way.
- **"The deleted assertion was unreachable."** **Refuted** — see below. This one I did check.

---

### The false-current risk, specifically

**A `current` verdict is byte-truth today, and structurally so.** `classify` returns `current` only on `mine == current`, and I verified all 49 published `current` values equal the on-disk byte digest. There is no path by which a byte-different SKILL.md is called current. Probes: one byte appended → `DIVERGED` rc 1; `version:` hand-edited to garbage → still `current`, correctly, because the line is elided before hashing. The stamp cannot be used to fake currency in either direction. That is the central claim and it holds.

**The risk is not false `current`. It is exit 0 on a stale copy, and there are three routes.**

**(1) History order — unguarded, and it is the ungated half.** `classify` computes `STALE n` from `history.index()`. Real-data probe: I recovered `api-contract-enforcement @ a8a84b90d865` out of git history and planted it as a subscriber. Shipped index → `STALE 2`, rc 1. Same index with **only that slug's history reversed** → `unpublished`, **rc 0**. A genuinely two-versions-behind copy is told nothing is wrong. Then the mutation: `reverse=True` at all three sort sites in `gen_skill_versions.py` (`:150`, `:156`, `:228`), grep-confirmed 3 occurrences → **361/361 green, exit 0**. Restored → green. Control on the same subset (drop the elision in `skill_version.digest`) → **79 failed, exit 1**, so that green is a real green. And `grep -rniE "chronolog|newest|oldest|order"` over both relevant test files returns 0 lines, exit 1 (positive control, same pipeline: `history` → 21 lines, exit 0). The index's own `verification` block declares 102 of 151 rows **NOT GATED** — that honesty is to the PR's credit, but the ungated half is exactly the half every `STALE` verdict is computed from.

**(2) `unpublished` and `retired` exit 0.** `skills_current.py:205` returns non-zero only for `STALE`/`DIVERGED`, while the docstring's exit table says 0 means "current (or local, or mirrored)". Demonstrated above: a two-versions-behind copy reading `unpublished` exits 0. `retired` is fixture-only today (0 of 49 entries have `current: null`).

**(3) The digest covers SKILL.md and nothing else.** Synthetic subscriber, catalog slug, local index: I deleted an entire `references/` directory → `3 skill(s) checked: 3 current`, **rc 0**. Control: one byte appended to that SKILL.md → `DIVERGED`, rc 1. Live scope, measured: the catalog has **0** non-`SKILL.md` files today (control: 49 `SKILL.md`), and exactly **one** installed skill dir anywhere in the consumer set carries more — `arlyn-working/empirical-design-principles`, 113 KB across three `references/*.md` — and it is not a catalog slug. So zero live false-currents, and a guaranteed one the day #237's multi-file skills reach the catalog. Issue #237 currently says this lane is "unaffected"; that is wrong and should be corrected in the issue.

**One thing that is *not* a false-current risk, despite looking like one.** `validate_skills.py` digests decoded text; every other tool digests bytes. That diverges only on a lone `\r` — verified: lone-CR file gives `c09cd7f5b304` (bytes) vs `4d210a62a876` (text), while LF and CRLF controls agree exactly. Both directions fail *loud*, never silent, so it cannot produce a false current. It can produce a **deadlock**: I injected a lone CR, ran `stamp_versions.py --write` (rc 0), then `--check` → `OK: 49 stamps match their file's content digest`, rc 0 — and `validate_skills.py` → rc 1, "`version` claims 82c1a4a270b0 but this file's content digest is e606dd248a0a … Run `python3 .github/scripts/stamp_versions.py --write`". The remedy is the command that just ran. Fix is one line.

---

### Fix before merge

1. **`validate_skills.py:162`** — `skill_md.read_text(encoding="utf-8")` then `:227` `content.encode("utf-8")`. Use `skill_md.read_bytes()`. This is a second owner of "the one rule" the whole design rests on, in the file that gates it. No test pins the current behaviour, so the swap is safe.
2. **`skills_current.py:148` / `:157`** — the `repo-mirrored:` check sits *below* `mine in history and current in history`, so it is unreachable for the case it was written for. Probe: `logmind`'s authoring repo holding the previously-published digest → `STALE 1` **plus `update: npx skills add …`**, rc 1 — an instruction to overwrite an authoring repo's own source from a mirror the docstring says is "not the authority". Control: the same repo with an unpublished local edit → `mirrored`, rc 0. Latent today (measured: 0 live occurrences across `tokenomics` and `logmind`), but it is the normal state of an authoring repo between publishes. Move the `repo-mirrored:` branch above the history branch.
3. **`skills_current.py:205` + the docstring exit table** — make `unpublished`/`retired` agree with what the code returns, or make the code agree with the table. Either, but not neither: this is the last remaining way to print a benign word and exit 0 over a copy that is behind.
4. **A test that pins chronological history order.** One assertion. Today three sort-key mutations survive green.
5. **`validate_skills.py:478`** — `if versions_path.exists():` silently disarms the whole index-currency gate. Probe: delete `docs/skill-versions.json` → `OK: 49 skills validated cleanly.`, **exit 0**. Control: index present with one `current` corrupted → exit 1, 2 errors. `skills_current.py` exits 2 on an unreadable index and its docstring argues at length for why; the catalog-side gate should not be the one that reports success from an absence.

**Follow-ups, not merge blockers:** README's "Two gates" and its add-a-skill recipe are untouched (`grep -cE '^(README|AGENTS)\.md$'` over the 64 changed files → 0, exit 1; control `^docs/` → 5, exit 0) and the recipe now produces a red PR. The `enumeration` provenance block does not reproduce in a fresh clone, and one frozen row is unresolvable there: `9e74682` lives only on `origin/fix/a11y-2588-conjunction`, which `git ls-remote` shows deleted upstream (0 lines; control `dev` → 1). All 44 distinct history shas resolve in a local clone with stale tracking refs, which is why the generator produced them at all. Append-only means that row never heals.

---

### Credit where it is earned

**The footer refusal is the best decision in the PR, and it is measured, not asserted.** The route home lives in the frontmatter because the size gate charges the body only, and `test-discipline` has **1 byte** of headroom under the 8192 cap. A body footer would have been unshippable on arrival. I re-derived the cap, the measurement site, and the headroom independently.

**The fixed point is real and it is why this PR does not invalidate the installed base.** Deleting the `version:` line from all 49 head files reproduces `origin/dev` byte-for-byte. Consequence I measured: **0 of 92** installed copies across the consumer set carry a `version:` line (control: 49 of 49 catalog head files do) — and every one of them still verifies, because the elision makes the stamp invisible to the digest. Adding a line to 49 files without moving a single digest is the design working exactly as documented.

**The zero-coverage refusals are real and they are the reason the verdict is "with fix" rather than "do not merge."** `stamp_versions.py` refuses an empty walk; `validate_skills.py`'s `coverage_errors` asserts `validated == on-disk count`; a sibling lens drove eight independent routes to zero coverage in `skills_current.py` and every one exits 2, never 0. The failure this mechanism exists to remove has genuinely been removed at the zero end. The gap is at the *partial* end: I built a tree with 5 skill dirs of which 2 were unreadable (empty dir, dangling symlink) and got stdout **byte-identical** to the 3-dir baseline, exit 0 — no warning, no count of what was skipped. Measured at 0 occurrences across all 12 live repos today, which is why it is a follow-up and not item 6.

**One claim I have to withdraw credit for.** `stamp_versions.py` deletes a `stamped_value(new) == digest(new)` assertion with the comment "a mutation run showed no sabotage could reach it without tripping this line first," and adds "An assertion nothing can falsify is not a check." I falsified that with my own probe: a **persistent** unquoting regression in `skill_version.version_line` is accepted by every surviving assertion — `restamp` returns cleanly, `stamped_value(new)` is `None`, `digest(new)` is `e606dd248a0a`, so the deleted assertion evaluates **False**. Controls both directions: unmodified → True; the builder's own **first-call-only** sabotage shape → `AssertionError: not idempotent`. The mutation experiment that justified the deletion was under-powered by exactly the shape it used. Consequence is bounded — `validate_skills.py` does catch an unquoted stamp (rc 1, correct message, verified) — so this is defence-in-depth, not the only line. But the source comment currently states something false in a repo whose rule is that a doc is a source of truth. Restore the assertion or correct the comment.

---

### What I could not check, so "clean" is bounded to this

- **The real `npx skills add` round trip with a stamp present.** 0 of 92 installed copies carry a `version:` line today, so the exact new frontmatter line has never been through the installer in any live consumer. The compatibility argument in the PR body is an argument from protocol SPEC §2.1, not a measurement. This is the largest remaining unknown.
- **The live HTTPS path.** `docs/skill-versions.json` is not on `main` yet, so every run I drove used `--index <local file>`. Redirects, a 200 carrying an HTML error page, and a truncated body (`http.client.IncompleteRead` is an `HTTPException`, not an `OSError`, so it escapes `load_index`'s except clause) are untested by me and by the suite.
- **CI on the real runner.** Read statically only. Note `tests/test_skills_current.py:162` does a live DNS lookup for a `.invalid` host and will flake behind a captive resolver.
- **The 102 ungated history rows.** I confirmed the *disclosure* and confirmed the order is correct today on real data for the one slug I recovered blobs for. I did not re-derive all 102.
- **Whether the 12 repos I read are the 11 the builder audited.** I measured today's trees on this machine; the builder's set at audit time is not recoverable from here.

thrillmot:
## All five fixes applied, plus the restore-or-correct item — pushed as `09fb415`

Each carries a mutation proof (break it → confirm red → restore → confirm green), and I re-verified
the load-bearing ones myself rather than accepting the report.

| # | fix | mutation proof |
|---|---|---|
| 1 | `validate_skills.py` reads bytes once and decodes from them; the later re-encode is deleted | reverted to `read_text()` → both new tests RED (CR-only file silently passed; lone-CR body threw a spurious stale-stamp error) |
| 2 | `repo-mirrored:` moved above the history branch | reverted the order → `STALE 1` instead of `mirrored`, exactly the finding |
| 3 | `retired` joins the nonzero-exit set; docstring states the full 0/1 partition | reverted the return line → retired test RED, `unpublished` control stayed green |
| 4 | test pinning ascending chronological history order | `reverse=True` at all three sort sites, grep-confirmed → **only the new test** went red, 10 others green — reproducing your "361/361 green" |
| 5 | the `if versions_path.exists():` guard is gone; a missing index is an error | reverted → `test_an_absent_index_is_rejected` RED |

**Verified by me, independently of the fix report:**

```
raw = skill_md.read_bytes()      1        raw = content.encode(...)     0
if versions_path.exists(...)     0        'not a fixed point'           1
CONTROL 'def run'                1

validate_skills.py               OK: 49 skills validated cleanly.       exit 0
stamp_versions.py --check        OK: 49 stamps match ...                exit 0
scratch tree, index deleted      exit 1        <- was 'OK: 49 ... ' exit 0
```

### On the deleted assertion — you were right, and it was proved before it was restored

The lane reproduced your refutation before touching the code: monkeypatching `version_line` to drop
quotes **persistently**, `restamp` returned cleanly with `stamped_value(new) = None` against
`digest(new) = f8259af7c464`, so the deleted assertion would have evaluated False. The assertion is
restored and the false comment rewritten, plus a regression that patches `version_line` rather than
`stamp` — so both calls inside `restamp` are corrupted identically and idempotence is blind to it,
which is exactly the shape the original mutation missed.

### Fix 5 had a blast radius worth naming

93 unrelated tests assumed a missing index was silently tolerated. Rather than patch 93 assertions,
`conftest.py`'s `SkillTree.validate()` now auto-provisions a matching index computed from current
bytes, only when a test hasn't written its own — so per-skill-rule tests stay decoupled from the
currency gate, while `test_an_absent_index_is_rejected` bypasses that convenience and calls
`validate_skills.run()` directly. Fixing the class rather than the call sites.

### Suite

**830 passed, exit 0** (read from the process, not through a pipe) — up from 824; the six new tests
are the regressions above.

### Two of your follow-ups, tracked rather than silently dropped

- **#237 corrected.** You are right that it says this lane is "unaffected" by the one-file digest blind
  spot. It is not — the digest covers `SKILL.md` only, so deleting a whole `references/` directory
  reports `current`. Zero live false-currents today, guaranteed one the day #233's multi-file skills
  land. I am fixing that issue's text.
- **README's "Two gates" and its add-a-skill recipe** — now produces a red PR, since a new skill needs
  `gen_skill_versions.py --write`. Not in this push; filing it rather than leaving it in a comment.

The `enumeration` provenance row frozen at `9e74682` (a branch since deleted upstream) I am leaving as
you described it — append-only means it never heals, and that is a property of the design worth
carrying openly rather than papering over.

thrillmot:
## Verdict: MERGE — #238 `feat/per-skill-versioning` → `dev`

All five named fixes are real, present in the code, and load-bearing (each one control-broken to red). The union merge of `validate_skills.py` is structurally clean and, more importantly, behaviorally correct — verified independently, not just by re-reading the "already fixed" commit messages. Two full local suite runs are green: **970 passed, 0 failed, exit code 0** (explicitly captured, not inferred). One non-blocking, narrow latent-fragility finding in the *test-only* conftest ownership heuristic is reported below for a follow-up hardening note; it does not affect the current suite's correctness and does not block this merge.

Diffed throughout against `origin/dev` (`git diff origin/dev...origin/feat/per-skill-versioning`), never `main`, from a read-only `git archive` of the branch — nothing was written back to the real repo.

### 1. The five fixes — each verified present AND control-broken to red

| # | Fix | Where | Control-break result |
|---|---|---|---|
| 1 | `validate_skills.py` reads `skill_md.read_bytes()` then decodes, instead of `read_text()` | `.github/scripts/validate_skills.py:221-222` | Reverted to `read_text()` → `test_a_cr_only_file_fails_missing_frontmatter` goes red (asserts "missing YAML frontmatter", got a wrong currency-gate error instead — the byte/text divergence the fix exists to close) |
| 2 | `repo-mirrored:` check moved above the `history` branch | `skills_current.py:150-158`, above the `mine in history and current in history` check | Reverted (moved back below) → `test_a_repo_mirrored_skill_already_in_history_is_still_mirrored_not_stale` goes red: verdict flips from `mirrored` to `STALE 1`, exactly the defect the commit describes |
| 3 | `retired` joins the nonzero-exit set | `skills_current.py:214` — `return 1 if tally.get("STALE") or tally.get("DIVERGED") or tally.get("retired") else 0` | Reverted → `test_a_retired_skill_never_produces_an_install_instruction` goes red: `returncode == 0` instead of `1` |
| 4 | History rows sorted ascending by full timestamp, pinned by a dedicated test | `gen_skill_versions.py:156,228` (`sorted(..., key=lambda r: r[3])`) | Applied `reverse=True` → `test_history_rows_are_ascending_chronological_order` goes red (rows come back newest-first) |
| 5 | A missing/unreadable `docs/skill-versions.json` is an error, not silently skipped | `validate_skills.py:696-700` | Removed the `errors.append(...)` on the except branch → `test_an_absent_index_is_rejected` goes red: `only(errors)` gets `[]` instead of one error |

Each revert reproduced the *specific* defect the fix's own commit message names — not a coincidental, differently-shaped failure — which is what makes these real regression guards rather than assertions that happen to pass.

### 2. The union-merge risk — no third bug found

- `grep -c '^<<<<<<<\|^=======\|^>>>>>>>'` over `validate_skills.py`: zero conflict markers.
- `ast.parse` on the file: parses clean.
- No duplicate `def run` / `def directory_errors` / `def readme_errors` / `def _is_superseded` — each appears exactly once.
- Both gates fire in the merged `run()`: `errors.extend(directory_errors(...))`, `errors.extend(readme_errors(...))`, then the index-currency block — in that order, no interleaving bugs.
- **Independently recomputed all 51 digests** in `docs/skill-versions.json` against the real `skills/*/SKILL.md` bytes using `skill_version.digest()` directly (not through the gate under test): **0 mismatches, 51/51 checked**.
- Tree/index membership: `skills/*/SKILL.md` slugs vs `docs/skill-versions.json["skills"]` keys — symmetric difference is empty (51 in tree, 51 in index, exact match).
- The two previously-fixed sub-bugs from commit `a57e009` ("Make the two gates' test fixtures survive each other") were re-confirmed present in the current tree: the mutation harness (`test_skill_directory_mutations.py`) copies `skill_version.py` and the real `docs/skill-versions.json` into its scratch tree (`_scratch()`, lines 394/403); and the conftest ownership-by-content mechanism (below) is in place.
- **`test_control_the_unmutated_tree_is_green`** (#243's mutation-harness control) was run in isolation and is genuinely green — the mutation suite's reds elsewhere are therefore evidence about the mutations, not about a broken harness.

### 3. Conftest ownership-by-content — attacked, one narrow finding

Wrote three adversarial stress tests against `SkillTree.validate()`'s byte-identity ownership rule (`tests/conftest.py`):

1. **Can the scaffold's own index go stale?** Skill mutated directly on disk between two `validate()` calls, scaffold owning the index both times → second call recomputes fresh, no spurious currency error. **PASSED** — no staleness.
2. **Can a test's own index survive an unrelated later skill addition?** Test writes a custom (intentionally incomplete) index once, then adds a skill via `.skill()` without touching the index, then calls `validate()` again → the custom index is untouched, and the expected `skills.beta is missing` error fires. **PASSED** — ownership correctly not reclaimed.
3. **Coincidental byte-identity reclaim.** Scaffold writes and owns the index first (`self._auto_index` set). Test then hand-rewrites the index to content that is *byte-identical* to what the scaffold already wrote (a plausible single-skill collision, since both `index_for()`'s and the scaffold's JSON serialize identically for one entry). A subsequent `.skill()` addition + `validate()` call: the scaffold silently reclaims ownership (content still equals `self._auto_index`) and overwrites with fresh, correct data — discarding what the test believed it had pinned, with **no error surfaced**. **This assertion failed** — i.e., the collision is real and reproducible.

This is a genuine, narrow edge case in the "ownership by content, not by flag" design: content-equality is not the same as authorship, and the two can coincide. Two mitigating facts keep it non-blocking: (a) I checked every current caller of `index_for()` in `test_validate_skills_version.py` and none call `tree.validate()` before writing their custom index — `self._auto_index` is always `None` going in, so the collision path is inert in the actual suite today; (b) when it does fire, the scaffold's overwrite is *correct* data (freshly recomputed from disk), so it can only cause a test relying on a deliberately-stale index to get an unexpected (loud) assertion failure later, never a silent false-pass. Recommend a follow-up: track ownership with an explicit sentinel/flag rather than content equality, to remove the coincidence risk entirely — but nothing here blocks this merge.

### 4. Full suite — two independent runs, exit code captured

Run 1 (`pytest tests/ -q`, background, output to log): `970 passed in 260.21s` — log grepped for `FAIL|ERROR|error`, zero hits across all 15 lines of output.

Run 2, rerun specifically to capture the process exit code into a variable/file **before** any pipe (per the discipline this repo has been burned by): `bash -c "pytest tests/ -q > log 2>&1; echo $? > log.rc"` → `970 passed in 263.55s`, **`EXIT=0`** read from the `.rc` file.

`pytest --collect-only` on the same tree: **970 tests collected**, matching the 970 passed — confirms nothing was silently deselected.

Targeted version-suite subset (`test_skill_directory_mutations.py test_stamp_versions.py test_gen_skill_versions.py test_skill_version.py test_validate_skills_version.py test_skills_current.py`): **305 passed**, run separately for faster signal, also clean.

### 5. Corroborating, not primary, evidence

`gh api .../commits/<head-sha>/check-runs` on `ab55777` (current branch tip): `validate skills`, `check prose retention`, `check-links`, `check-decisions`, `check-derived-docs`, `test` all `success`; `clud-bug-review` is `neutral` ("review unavailable") — consistent with the repo's documented ongoing outage since #188, not a new problem, and exactly why this panel review is the actual gate here per `AGENTS.md`.

### What I re-derived vs. took on report

Re-derived from scratch: all five control-breaks, the 51/51 digest cross-check, the conflict-marker/parse/duplicate-def scan, both full-suite runs with the second run's exit code, the mutation-harness control test, and the three conftest stress tests. Taken on report (and separately corroborated via GitHub's check-runs API, not just re-read): that CI is green on the branch tip.

Issue #234, opened 2026-08-17
check-links is blind to skills/ — and it is now a required check


thrillmot:
`check-links` reports clean on `skills/` because it never looks there. Mutation-tested, both directions:

```
$ printf '\n[deliberately broken](../zzz-not-a-real-skill/SKILL.md)\n' \
    >> skills/proving-an-absence/SKILL.md
$ logmind check-links
All markdown links resolve and no orphans found.
exit 0

$ printf '\n[control broken](./zzz-nope.md)\n' \
    >> docs/integrating-with-agent-skills.md
$ logmind check-links
Broken links (1):
  - docs/integrating-with-agent-skills.md: missing -> ./zzz-nope.md
```

The control is the point: the checker works, it is simply blind to one directory — and it is the directory this repository exists for.

## Why this is worse than having no check

**It is now a required status check.** It was added to `reporulez-default` today alongside `validate skills` and `test`. So the repo has a merge gate that reports green over 54 skills it has never read, and the green is indistinguishable from the green it would give if every link resolved.

**Skills are where the links are.** The whole L0/L1/L2 model is cross-references — dispatchers routing to primitives, primitives routing back. Those are exactly the links nothing verifies.

**We have already shipped dead ones.** Three cross-references pointed at `diagnosing-a-blocked-deploy` after it was deleted, in a consumer repo. They were found by a hand-rolled Python resolver, not by this gate. Anything similar inside `skills/` here would have merged silently.

## What I would want, offered not prescribed

Whatever walks `docs/` should walk `skills/` too, including the new `references/` subdirectory. Two wrinkles worth deciding deliberately rather than discovering:

1. **Sibling-skill links are relative** (`../apca-contrast/SKILL.md`) and resolve on disk. Straightforward.
2. **Links into `references/` are absolute GitHub URLs** — deliberately, because a relative one 404s on skills.sh, which renders `SKILL.md` alone. A link checker that only resolves paths will not see these, and one that fetches URLs will 404 on any reference file not yet on `main`. That second case is real right now: PR #233 adds three, and their absolute URLs correctly 404 until it merges. A checker that fetches would block its own PR.

**And mutation-test whatever lands.** This gate has presumably reported clean on `skills/` since it was written, and nobody could tell — which is the argument for the check having to fail on demand before anyone trusts it. Same rule the catalog applies to its own guards.

Found while preparing #233; the PR body carries the same finding so nobody cites a green check on it.

thrillmot:
[...]

Issue #235, opened 2026-08-17
frontend-a11y overstates SC 2.5.8: four of five exceptions unowned, and the box test is stricter than the spec


thrillmot:
`frontend-a11y` renders SC 2.5.8 more strictly than the spec does, and names one of its five
exceptions. Both produce **false findings against a client's code** — the failure direction that costs
a lane a day arguing with a wrong verdict.

Surfaced by the normative-correctness lens reviewing #232 (which fixes a different, real defect in the
same sentence and should merge on its own terms). None of this is introduced by that PR.

### 1. The Inline exception is missing catalog-wide — the costly one

SC 2.5.8 has five exceptions. The skill names **Spacing** only, and with the definite article ("the
spacing exception"), which reads as though it were the only one.

**Inline** exempts targets "in a sentence or [whose] size is otherwise constrained by the line-height
of non-target text". Inline links in body copy are undersized by construction and are *exempt*. An
agent running this skill against a rendered page will report every one of them as an SC 2.5.8 AA
failure. For a lens whose whole job is judging rendered surfaces, that is not an edge case — it is
most pages.

Measured, with controls:

```
skills/frontend-a11y/SKILL.md          "spacing exception"  1     "2.5.8"  3      <- probe works
skills/frontend-a11y/SKILL.md          Inline|Equivalent|Essential|User Agent Control|bounding box  0
skills/component-sizing-principles/    same terms  0        "24 CSS px"  5        <- probe works
skills/spacing-system/                 same terms  0
```

Zero coverage in the three skills that could plausibly own it.

### 2. "another target's box" is stricter than the normative text

`skills/frontend-a11y/SKILL.md:66` tests the circle against "another target's **box**". The spec tests
against "**another target**" — "bounding box" appears in the normative text *only* to locate the
circle's centre, and Understanding 2.5.8 is explicit that the two differ for non-rectangular targets:

> If the target is not rectangular — for instance, if the target is clipped, has rounded corners, or
> if it's a more complex clickable SVG shape — we need to first determine the bounding box

**Failure case:** an undersized target whose circle clears a rounded-corner neighbour's actual region
but clips its bounding box. Spec says pass; the skill says fail.

The fix is shorter than what's there, which matters at 64 bytes of headroom — it **returns 6 bytes**:

```
must miss **another target** — not its circle — and another undersized target's circle.
```

### 3. "a 24 px circle" is diameter/radius ambiguous

[...]

PR #240, opened 2026-08-18
composing-a-screen: the prescriptive design spine the catalog never had

thrillmot:
The catalog had no skill that tells an agent how to **make** a screen. This adds one.

## The gap, measured

Against the seven basics a free [Figma resource page](https://www.figma.com/resource-library/ui-design-principles/) covers, this 49-skill design-heavy catalog scored:

| principle | skills covering it |
|---|---|
| contrast | **12** |
| accessibility | **12** |
| consistency | 3 |
| hierarchy | 1 |
| alignment | 1 |
| **proximity** | **0** |
| **progressive disclosure** | **0** |

Controls: `zorkmid` → 0, `design` → 24, so the zeros discriminate.

Twelve skills deep on contrast, nothing on four of the seven. The catalog grew out of `design-critic` needing *citable* lenses, so depth accumulated exactly where a finding can be argued numerically and nowhere near where design judgement lives. Every design skill is review-framed. A ten-second Google search covered our four zeros.

## What this is

`composing-a-screen` — nine steps in dependency order, each handing back a default or a forbidden move rather than a description. Rank → group → encode → space → lay out → conventionalise → target → clear the floors → defer.

It **routes down** rather than restating: values come from `spacing-system`, `type-scale`, `component-sizing-principles`; thresholds from `frontend-a11y`, `apca-contrast`, `wcag-contrast`; the visual bar from `designing-elite-ui`, which it names as REQUIRED BACKGROUND.

Where the honest answer is judgement it says so — steps 1 and 5 are explicitly marked *(Judgement, no threshold.)* rather than inventing precision. The Sources block carries its own scope condition:

> the grouping ratio is dot lattices transferred to UI, not measured on it — which is why 2 is a default, not a discovery

and closes with a **Do not repeat** naming two things it refuses to assert (modular type-scale ratios; the uncited "80% of customers lost" for Jakob's Law).

## The panel caught three defects that would have made screens worse

Not stylistic — wrong in the direction that damages output:

1. **Fitts's Law stated backwards** — movement time "falls" with distance ÷ size.
2. **The hierarchy rule forbade redundant coding**, which is *faster*, not slower.
3. **The proximity metric was `longer ÷ shorter`**, which scores an inverted layout as ideal.

Then its own severability grep caught a fourth: the body contained **zero** instances of the exact phrase "progressive disclosure" (the step was titled *"Decide what waits behind a click"*), so the skill would have scored 0 on the very census query that established the gap — **the same defect the panel confirmed in PR #233**. Retitled; same for "accessibility", which had been compressed to "a11y".

## Verification

Four control probes, each seen to fail and recover:

| probe | injected | result |
|---|---|---|
| size gate | +143 B | `over the 8192-byte limit by 72` exit 1 → restore → `OK: 50` exit 0 |
| placement map | entry deleted | `missing an entry for: composing-a-screen` exit 1 → restore → exit 0 |
| link checker | `../visual-polish-NOPE/SKILL.md` | `BROKEN: 1` exit 1 → restore → `BROKEN: 0` exit 0 |
| slug checker | `` `zorkmid-scale` `` | `NOT-A-SKILL zorkmid-scale` exit 1 → restore → exit 0 |

Boundary-tested against the validator's own arithmetic: +95 B → `body is 8193 bytes, over ... by 1`. So **8,097 bytes, 95 headroom** is the gate's number, not mine.

Links verified by hand — 5 relative, all resolve; 8 backticked slugs, all real directories — because `check-links` does not read skill bodies (#234).

**`pytest` was unavailable in the build sandbox and is reported as unrun, not as passing.** CI runs it here.

## Placement

`opt-in`, diverging from the other three L1 dispatchers' `catalog-only`. Their note explains their choice — *"a repo consumes what it routes to, not the dispatcher itself"* — and that reasoning does not transfer: this one also **carries** the principles nothing else owns, so `catalog-only` would make a subscribing repo raise a placement verdict for installing the only copy of that content. Deliberately absent from `census_counters`' `STRUCTURAL_L0` frozenset, so zero subscriptions reads as a real gap rather than by-design.

## What it does not claim

It produces a correctly **organised** screen, not a beautiful one. It says nothing about colour, type personality, motion, or restraint, and cannot inside 8,192 bytes — that is `designing-elite-ui`'s job, which is why the REQUIRED BACKGROUND link is load-bearing rather than decorative.

**Whether an agent actually designs better with it is untested.** The behavioural diff — run an agent through a screen with and without this file — has not been done. Treat "it helps an agent design well" as a hypothesis this file makes testable, not a result.

## Reviewer notes

- `.gitignore` gains `.playwright-mcp/` — browser-automation debris found untracked in the tree; it should never be committable by anyone.
- My first draft of the `placement-map.json` edit reserialised and alphabetised the whole file (322-line diff for one entry, destroying a curated family grouping). Caught in self-review and redone as a 6-line textual insertion; the original 49 entries' order is preserved, verified programmatically.
- `docs/file-structure.md` is **deliberately not updated** — regenerating it adds `.claude/worktrees` and `scheduled_tasks.lock`, both `tracked-files=0, ignored=YES`. That is #221, and I am not committing gitignored paths into a tracked doc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 # python
 __pycache__/
 *.pyc
+
+# Playwright MCP session debris (console logs, page snapshots) — never repo content
+.playwright-mcp/
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ The design catalog is organized in three layers (locked by the CDO; executed in 
 | [`designing-a-design-system`](skills/designing-a-design-system/SKILL.md) | Dispatcher for **building or extending** a design system — routes through naming → color → non-color families → format/versioning → the elite bar → testing, in build order. |
 | [`reviewing-design-work`](skills/reviewing-design-work/SKILL.md) | Dispatcher for **reviewing or critiquing** design output — ordered lenses (code rules → rendered-surface lenses → the opinion bar) with routing rules for when the browser-driven design-critic pass fires. |
 | [`consuming-a-design-system`](skills/consuming-a-design-system/SKILL.md) | Dispatcher for **using** a design system in a product — token discipline, composition rules, DTCG install patterns, SemVer-aware upgrades, extend-vs-fork. |
+| [`composing-a-screen`](skills/composing-a-screen/SKILL.md) | Entry point for **making** a screen — the composition sequence in dependency order (rank → group → encode → space → lay out → conventionalise → target → clear the floors → defer). Carries the principles nothing else owns: hierarchy, proximity as a between-to-within ratio, measure and column count, alignment axes, progressive disclosure, Jakob's and Fitts's. |
 
 ### L0 — universal primitives
 
--- a/docs/decisions-branches/feat__composing-a-screen.md
+++ b/docs/decisions-branches/feat__composing-a-screen.md
@@ -1,0 +1,13 @@
+← back to [docs/timeline.md](../timeline.md)
+
+## 2026-08-18 02:04 - Add composing-a-screen: the prescriptive spine the design catalog never had
+
+**Reasoning:** Measured coverage against the seven basics a free Figma resource page covers: contrast 12 skills, accessibility 12, consistency 3, hierarchy 1, alignment 1, proximity 0, progressive disclosure 0 (control: zorkmid 0, design 24). The catalog grew out of design-critic needing citable lenses, so depth accumulated exactly where a finding can be argued numerically and nowhere near where design judgement lives. Every design skill was review-framed; nothing told an agent how to compose a screen. The CEO's test was that a ten-second Google search covered our four zeros.
+
+**Alternatives considered:** Extend designing-elite-ui, or grow empirical-design-principles into the spine. Rejected: the first is the visual bar and this is structure, and the second explicitly scopes itself to falsifiable prediction and disclaims usability heuristics including progressive disclosure. A third option, doing nothing on the grounds that the depth already exists, fails because depth an agent cannot route to is not coverage.
+
+**Implications:**
+- L1 and opt-in, diverging from the other three dispatchers' catalog-only: catalog-only makes a subscribing repo raise a placement verdict, which would make this skill un-installable by the repos it is written for. Deliberately absent from census_counters STRUCTURAL_L0, so zero subscriptions reads as a real gap rather than by-design. A refute-first panel caught three defects that would have shipped guidance making screens worse: Fitts stated backwards, a hierarchy rule forbidding redundant coding, and a proximity metric that scored an inverted layout as ideal. The skill's own severability grep then caught that the body contained zero instances of 'progressive disclosure' — it would have scored 0 on the very census query that established the gap, the same defect PR #233 has. Body 8097 bytes, 95 headroom. Whether an agent actually designs better with it is untested and stated as a hypothesis, not a result; the behavioural diff has not been run.
+
+---
+
--- a/docs/file-structure.md
+++ b/docs/file-structure.md
@@ -7,8 +7,6 @@ agent-skills
 ├── .claude
 │   ├── commands
 │   ├── skills
-│   ├── worktrees
-│   ├── scheduled_tasks.lock
 │   └── settings.json
 ├── .github
 │   ├── ISSUE_TEMPLATE
--- a/docs/file-structure.md
+++ b/docs/file-structure.md
@@ -92,9 +92,6 @@ agent-skills
 ├── .cursorrules
 ├── .gitattributes
 ├── .gitignore
-├── .merge_file_Ggwrba
-├── .merge_file_OVMET7
-├── .merge_file_QRUAns
 ├── AGENTS.md
 ├── CLAUDE.md
 ├── LICENSE
--- a/docs/file-structure.md
+++ b/docs/file-structure.md
@@ -7,6 +7,8 @@ agent-skills
 ├── .claude
 │   ├── commands
 │   ├── skills
+│   ├── worktrees
+│   ├── scheduled_tasks.lock
 │   └── settings.json
 ├── .github
 │   ├── ISSUE_TEMPLATE
@@ -33,6 +35,7 @@ agent-skills
 │   ├── clud-bug-collaboration
 │   ├── component-sizing
 │   ├── component-sizing-principles
+│   ├── composing-a-screen
 │   ├── consuming-a-design-system
 │   ├── conventions-template
 │   ├── critical-issues-only
@@ -89,6 +92,9 @@ agent-skills
 ├── .cursorrules
 ├── .gitattributes
 ├── .gitignore
+├── .merge_file_Ggwrba
+├── .merge_file_OVMET7
+├── .merge_file_QRUAns
 ├── AGENTS.md
 ├── CLAUDE.md
 ├── LICENSE
--- a/docs/placement-map.json
+++ b/docs/placement-map.json
@@ -201,6 +201,8 @@
       "notes": "L1 dispatcher — entry point that routes to review lenses; a repo consumes what it routes to, not the dispatcher itself as a standing subscription"
     },
     "composing-a-screen": {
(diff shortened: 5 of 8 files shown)


thrillmot:
## Verdict: MERGE_WITH_FIX

PR #240 adds `composing-a-screen`, a genuinely strong, densely prescriptive design-composition
skill. Every numeric threshold and named law I checked against a primary source — Fitts's Law
direction, the Treisman & Gelade ~60-vs-~3 ms/item split, the 45–75 character measure, WCAG
SC 2.5.8's 24px floor, SC 1.4.8/F88's justify-text rule, the NN/g 2016 hamburger-menu study's
≥39%/n=179 figures, and Nielsen's verbatim 2000 Jakob's Law quote — checked out exactly. All
three previously-reported panel-caught bugs (Fitts backwards, hierarchy forbidding redundant
coding, an inverted-scores-as-ideal proximity metric) are verifiably absent from the current
file, not just claimed fixed. Both required gates pass cleanly when I ran them myself:
`validate_skills.py` (50 skills, exit 0) and `pytest tests/ -q` (570 passed, exit 0 — notably,
the author's own PR description says pytest was unavailable in their sandbox and reports it as
**unrun**, so this review is the first time that gate has actually been exercised against this
branch). Body is exactly 8,097 bytes against the 8,192 hard limit, matching the PR's own
claimed number. Every relative link and every backticked cross-reference resolves on disk
(`check-links` #234 never reads `skills/`, so I did this by hand with a known-broken control).
The `opt-in` vs the other three L1 dispatchers' `catalog-only` divergence holds up structurally
against every other entry in `placement-map.json`, not just the PR's own say-so.

**One thing blocks merge**, and it isn't in the new skill:

**Blocking — `skills/frontend-a11y/SKILL.md`, an unrelated one-word regression.** The PR bundles
an out-of-scope edit changing the SC 2.5.8 spacing-exception sentence from "...must miss another
target's box — not its circle — **or** another undersized target's circle" to "...**and**
another undersized target's circle." The W3C criterion text itself uses "or" ("the circles do
not intersect another target or the circle of another undersized target"), confirmed against
three independent WCAG-summary sources. This is a real correctness bug shipped into an
already-live, install-and-consume accessibility skill other agents depend on — small in size,
real in effect, and it has nothing to do with `composing-a-screen`. Restore "or" (or drop the
hunk) before merge.

**Non-blocking, but worth fixing in this PR or immediately after:**

1. **The PR's own motivating "proximity 0" claim is measurably false.** `grep -rli proximity
   skills/` on pre-PR `main` finds `visual-polish/SKILL.md`, which already carries a full
   paragraph ("Rhythm has to read as grouping") on the identical principle — relative distance,
   gap-inside-smaller-than-between, "Fix: two distinct steps" — and names "Gestalt proximity"
   explicitly. `composing-a-screen` §4 restates this same qualitative rule (its real
   contribution is the *numeric* ratio and citation, which visual-polish lacks and which is
   genuinely valuable), but neither file points at the other, so the catalog now states one
   principle in two places with no owner pointer in either direction — the asymmetric version of
   the catalog's own known defect #205. The honest fix is cheap: correct the coverage-table claim
   to "0 skills state a numeric ratio" (the real gap), and either in this PR or a fast-follow,
   trim `visual-polish` item 3 down to defer the ratio/reasoning to `composing-a-screen` §4 —
   the same pattern that skill already uses for `spacing-system` ("owns the scale; this lens
   owns whether it reads").
2. `measure` (the typographic noun promised in the frontmatter description) never appears in the
   body — only unrelated verb forms ("measured search...", "not measured on it..."). The concept
   is fully present in step 5 under different wording, so this is cosmetic, not a #233-style
   absent-concept defect.
3. The Verification section's `1–1.5 defect / 1.5–2 warning` tier boundary at 1.5 is never tied
   back to its source. Kubovy/Wagemans' actual dot-lattice threshold is ~1.5, not 2 — the skill
   is candid that "2 is a default, not a discovery," but never says 1.5 *is* the discovery. One
   clause would close the loop.
4. The PR description states "`docs/file-structure.md` is deliberately not updated," which
   looks contradicted by the diff. Traced it: a second commit
   (`chore: regen derived docs`, `logmind-auto-regen[bot]`) landed after the author wrote the
   description, via the repo's own `check-derived-docs` self-heal. Not a defect in the tree —
   just a stale description. One-line fix, not a blocker.

**What I verified myself vs. took on the PR's word:** I independently reran both required gates
(they hadn't actually been run — see above), reproduced the exact byte count, reproduced the
"8 backticked slugs, all real directories" claim down to the same count, and cross-checked the
`opt-in` placement reasoning against every sibling entry in `placement-map.json` rather than
just the note's own text. I did **not** access the primary Wagemans 2012 PDF at the cited page
(paywalled — corroborated the ~1.5 figure via a 2025 secondary source instead), did not
re-derive the full seven-principle coverage table beyond spot-checking hierarchy/alignment
(both plausible on inspection), and did not re-trigger CI directly (cross-checked
`gh pr view`'s live status instead, which shows every non-`clud-bug` check green and
`clud-bug-review` still `NEUTRAL`/unavailable on both attempts — independently confirming
AGENTS.md's documented outage and that no real review has touched this PR yet).

**Bottom line:** the new skill itself is ready. Pull the accidental `frontend-a11y` regression
out (or fix the word) and this merges clean; the proximity-overlap and description-accuracy
items are worth a follow-up but shouldn't hold the branch.

thrillmot:
## Overruling the blocking finding — it is a base-comparison artifact, and the logic is inverted

I authored this PR, so I am stating the evidence rather than the conclusion, and leaving the panel's
report above intact so the record shows what was overruled and why.

The finding is that `skills/frontend-a11y/SKILL.md:65` contains an out-of-scope correctness regression
— `or` changed to `and` in the SC 2.5.8 spacing exception.

### 1. That file is not in this PR

```
git diff --name-only origin/dev...origin/feat/composing-a-screen  -> frontend-a11y  0
git diff --name-only origin/main...origin/feat/composing-a-screen -> frontend-a11y  1
```

The change is **already on `dev`**, merged as #232 earlier today. The panel compared against `main`,
which does not yet carry it, so #232's diff appeared inside #240's. Nothing in this PR touches that
file — the seven files it does touch are `.gitignore`, `README.md`, the decision entry,
`docs/file-structure.md`, `docs/placement-map.json`, `docs/timeline.md`, and the new skill.

### 2. And `and` is correct, which is why #232 made the change

The spec is phrased as a **negation**:

> the circles do not intersect another target **or** the circle of another undersized target

The skill is phrased **positively**:

> a 24 px circle on each undersized target must **miss** another target's box … **and** another
> undersized target's circle

`¬(A ∨ B) ≡ ¬A ∧ ¬B`. "Do not intersect A or B" and "must miss A and B" are the same requirement.
Reading the spec's `or` straight across into a positive sentence inverts it — which is exactly the
defect #220 filed and #232 fixed.

This was settled against the Understanding document's own worked figures rather than its prose,
precisely because the sentence is easy to misread: **Figure 5** (20×20, zero gap) fails on
circle–circle alone; **Figure 7** (16px rows, 1px gap) fails on circle–target alone. Both sufficient
to fail independently ⇒ both required to pass ⇒ conjunction.

### 3. But there is a real finding buried in the wrong one, and I am taking it

**Three separate readers have now misread this one sentence in a single day** — #220's author, and
this panel. It is correct and it is genuinely hard to parse. That is a defect in the writing even
though it is not a defect in the content.

**#235** already proposes the rewrite that fixes it — dropping "another target's **box**" for
"another **target**", which is both closer to the spec and shorter. This adds a third data point that
it is worth doing, and I have said so on that issue.

### Verdict

Merging. The blocking finding does not describe this PR, and the change it describes is correct.
Everything else in the report — that all seven promised principles are delivered in the body, that
the three previously-caught bugs are verifiably absent rather than merely claimed fixed, and that
every threshold checked against a primary source held — I accept, and it is the substance of the
review.

PR #242, opened 2026-08-18
AGENTS.md: delete two recited counts rather than update them

### LINKED ISSUE #236: AGENTS.md recites '112 tests'; the suite is 570 — delete the count rather than update it
`AGENTS.md` tells every agent that the suite is 112 tests:

```
pytest tests/ -q                             # 112 tests; the gates' own regression guard
```

Measured on `dev` today, in an isolated checkout:

```
570 passed in 971.59s (0:16:11)
```

**570, not 112.** Surfaced by the gates lens reviewing #232, which ran the suite rather than citing it.

### The fix is to delete the count, not update it

`AGENTS.md` is the canonical instruction file — the first thing every agent reads. A baked-in number
there goes stale silently and is believed while it is wrong. Our own rule says so: *"Never recite a
number. Measure it, and show the command."* Replacing `112` with `570` re-arms the same trap for
whoever adds test 571.

Suggested:

```
pytest tests/ -q                             # the gates' own regression guard
```

The count is not load-bearing for any instruction in that file — nothing tells an agent to check it,
and no gate asserts it.

### Why this is worth a ticket rather than a drive-by edit

An agent that runs the suite, sees 570, and has been told to expect 112 has to decide whether it is
looking at a broken checkout. That is a real cost paid repeatedly and silently, and it is exactly the
"a mechanism that cannot report its own staleness" shape.

Low priority, trivial diff, but it is in the file with the widest blast radius in the repo.

thrillmot:
Closes #236.

Both numbers in `AGENTS.md` had gone stale, and the second one was mine.

| claim | measured now |
|---|---|
| `pytest tests/ -q  # 112 tests` | **570** on `dev`, **830** on `feat/per-skill-versioning` |
| "On every pull request since #188 — **twelve of them**" | **16** PRs after #188 |

**The test count is the sharper case: it differs by branch.** There is no value that is correct in a
file both branches share, so this is not a number that went stale — it is a number that cannot be
maintained. It moved twice in one day (570 → 824 → 830) as #238 added tests.

Deleting rather than updating, per this repo's own rule — *"Never recite a number. Measure it, and
show the command."* Replacing `112` with `830` re-arms the identical trap for whoever adds test 831.

Neither count was load-bearing: nothing instructs an agent to check either, and no gate asserts them.
The clud-bug section already carries the `gh api … check-runs` command that recovers its own figure,
which is what the count was standing in for.

**Left alone deliberately:** `_Installed at clud-bug v0.7.0-rc.20._` sits inside the marker block
regenerated by `clud-bug update`, so hand-editing it would be overwritten. That block is also where
the file itself says nothing measured belongs.

**Why raw git rather than `logmind log`:** the working tree also held `docs/file-structure.md`
restaged by the branch checkout (#221 — the generator writes gitignored paths into a tracked doc) and
an untracked `.playwright-mcp/`. `--stage all` sweeps the working tree and `--stage scoped` stages
only the decision file, so neither mode commits exactly this one file. Used an explicit pathspec
instead; the diff is two deletions of stale figures and carries no decision beyond the one this PR
body states.

*Aside worth noting: the "112" was independently rediscovered today by the lane fixing #238, which
measured 830 and flagged it unprompted. A number wrong enough to be caught twice in one day by
different readers is a good argument for it not existing.*


thrillmot:
## Verdict: MERGE

Three lines, both edits are correct, and nothing load-bearing was touched. One pre-existing, out-of-scope inaccuracy found nearby — worth a follow-up issue, not worth blocking this PR.

### What I re-measured (not took on report)

| Claim | PR says | I measured | Method |
|---|---|---|---|
| Test suite on `dev` | 570 | **570** | `pytest tests/ -q --collect-only` in an isolated `git archive` checkout of `origin/fix/agents-md-recited-counts` (which is `dev` + this 2-line diff) |
| Test suite on `feat/per-skill-versioning` | 830 | **830** | Same method, separate isolated archive of `origin/feat/per-skill-versioning` |
| Old count of PRs since #188 ("twelve") | superseded | consistent | See below — the old "twelve" figure checks out under one specific counting method |

Positive controls run first so these zeros/absences are trustworthy: grepped for a known real assertion (`SIZE_LIMIT = 8192`) and got a hit; grepped `"112"` unfiltered and got an unrelated but real hit (a PR-#112 reference in `dependabot-auto-merge.yml`). Both confirm the grep methodology isn't silently missing matches, so the "nothing load-bearing" finding below is trusted.

I did not personally watch a *full* (non-`--collect-only`) local suite finish green — issue #236 records that run at 971.59s (~16 min); my background run was still going (63%+, zero failures) when I wrote this. I'm relying on `--collect-only` counts (deterministic, reproduced twice) plus the PR's own CI `test` check reporting `SUCCESS` on the head SHA, plus issue #236's independently-dated "570 passed" measurement — three converging signals, not one recitation.

### 1. Prose reads correctly after both deletions

Both edited paragraphs read grammatically and make sense with the numbers gone:
- `pytest tests/ -q  # the gates' own regression guard` — clean, no dangling clause.
- `On every pull request since #188, clud-bug-review has concluded neutral...` — clean, no dangling em-dash.

### 2. Nothing load-bearing was deleted

Grepped `.github/workflows/*.yml` and `.github/scripts/` for any assertion tied to the test count or the PR count: none. `test.yml` runs `pytest tests/ -q` with no count check — it fails only on a failing test, never on a count. No script hardcodes "112", "570", "16", or "twelve" as a gate. The only other places "112 tests" appears are `docs/decisions-branches/docs__dev-branch-convention.md` and `.../feat__three-design-lenses-to-house-standard.md` — both are point-in-time decision-log narration (a record of what was true when written), not living-truth docs, so leaving them alone is correct per this repo's own doc philosophy ("reasoning and the story... go in the commit message").

`check-decisions` passing isn't the "batching dilutes" case the PR reasons about — I read `check-decisions.yml` directly: it categorically excludes `AGENTS.md` (and `docs/*`, `CLAUDE.md`, etc.) from the non-docs-line count, so a pure-`AGENTS.md` PR never needs a decision-log touch regardless of batching. Doesn't change the verdict, just a correction to the PR's own stated reasoning about why raw git was safe here.

### 3. File stays internally consistent — with one exception outside this PR's diff

`AGENTS.md` reads consistently after the edit, except for one thing this PR didn't touch: the immediately adjacent, un-edited sentence **"On every pull request since #188, `clud-bug-review` has concluded `neutral`... because no review is being produced"** is currently false. I pulled check-runs for every PR numbered >188 (18 total, `gh api .../check-runs`) and found **#218** concluded `failure` with a real body — *"clud-bug review — blocking critical findings"* — not the canned "unavailable" neutral. A review *was* produced, and strict mode *did* fail, contradicting both halves of that sentence. This predates PR #242 (the same absolute claim, minus the "twelve" count, was already on `dev`), and it's outside #236's stated scope (which was specifically the two recited counts). I'm not blocking on it, but flagging it as a nonblocking finding to file separately, per this repo's own rule that a finding outside a lane's own item gets filed, not folded into someone else's commit.

Side note on the "16" figure: it appears only in the PR *body* (a table), not in the shipped file — the diff removes "twelve" without replacing it with any number, which is the right move. I independently reproduced 16 as defensible under one specific method (all PRs numbered 189–243 = 18, minus #242 itself, minus #218 since it isn't neutral = 16), and the old "twelve" figure also checks out under a different, narrower method (merged-only PRs 189–232 = 13, minus #218 = 12). Both old and new figures happen to correctly exclude #218 — which makes the *retained prose's* unqualified "every... has concluded neutral" claim harder to defend, not easier.

### 4. The regenerated marker block was correctly left alone

`SKILL_GLOB_RE` in `check_prose_retention.py` scopes that gate to `skills/[^/]+/SKILL.md` only — it never touches `AGENTS.md` at all, so its `SUCCESS` here is a no-op signal, not a decision. Separately, confirmed via `diff dev...PR` that no line inside `<!-- clud-bug-start -->`…`<!-- clud-bug-end -->` (including `_Installed at clud-bug v0.7.0-rc.20._`) was touched. `.github/workflows/clud-bug-self-update.yml` runs `clud-bug update` weekly, which the workflow's own comment says re-renders exactly that template block — so editing it by hand would in fact be overwritten, as the PR claims. Both edits landed in the "### clud-bug is not reviewing here" section immediately below the block, which is explicitly where that block's own trailing sentence says measured content belongs.

### Minor, non-blocking nit

The PR body attributes the delete-not-update approach to *"this repo's own rule"*, quoting "Never recite a number. Measure it, and show the command." That exact phrase doesn't appear anywhere in this repo's docs (`grep -rni "recite a number" .` — empty). It traces to issue #236's body, which likely echoes the session's own standing instructions rather than a rule documented in this repository. Doesn't affect the shipped file, just a citation looseness in the PR narrative.

PR #243, opened 2026-08-18
finding-a-catalog-skill: the directory that travels with an install (#229)

### LINKED ISSUE #229: The catalog's directory does not travel: a subscriber gets N skill dirs and no map
The catalog has a directory. It is complete right now — every one of the 49 skills appears in the README table with a one-line purpose, verified:

```
skills named anywhere in README.md   49 of 49
missing                              0
CONTROL 'zzz-not-a-skill' found      False
```

Two problems with it, and the second is the one that costs something.

## 1. It is hand-kept and ungated

`docs/placement-map.json` is reconciled 1:1 against `skills/` directory names by `validate_skills.py` — that gate is what makes the placement-map guide's "the map is kept in sync" true rather than aspirational. The README table has no equivalent. Measured: zero references to `README` in `.github/workflows/*.yml` and in `validate_skills.py`.

So it is complete by diligence, and nothing keeps it that way. The next skill added is the one that breaks it, silently, in the file most readers meet first.

## 2. It does not travel to subscribers — and this is the real gap

A subscriber installs skills into `.claude/skills/`. The README stays here. Measured in `arlyn-working`, which subscribes to 23:

```
skill directories in .claude/skills/   29
index files (*.md at that level)        0
catalog README present                  no
```

**An agent working in a subscriber repo can list 29 directory names and learn nothing about what any of them are for.** To find out whether a sibling skill covers the thing it needs, it opens files one at a time — and the failure mode is not "reads slowly", it is "concludes the coverage is absent and writes the guidance again". That has already happened here twice this week:

- I proposed five studio operating skills; a survey found **twelve** further rules I was about to write already lived in this catalog or in superpowers — `subagents never commit`, `verify before claiming`, worktree isolation, refute-first panels, red-before-green, name-the-SHA. Six duplicates avoided, but only because that survey read all 49 skills in full. That is not a thing a normal task will do.
- #225 records the inverse: a consumer hit a real coverage gap from the inside and had no way to see it was a gap.

## The proposal

**A directory skill, generated from `skills/` and CI-gated.** Not a README section — a skill, because that is the artifact that travels. It installs alongside the others, so a subscriber repo carries its own map of what it subscribed to.

Contents: one row per skill — name, layer, and the one-line purpose that already exists in the README table. The source of truth stays `skills/*/SKILL.md`; the directory is derived, exactly as `roadmap.artifact.html` is derived and never hand-edited.

Gate it the way `placement-map.json` is gated: keys reconciled 1:1 against directory names, fail on mismatch. The mechanism is already written and already runs on every PR.

**Then every skill links up to it.** A footer line on each skill pointing at the directory, so an agent that has loaded exactly one skill can always navigate up and see the rest. That is the half that makes the directory reachable rather than merely present — a map nobody is holding is not a map.

## Branding, which rides in the same line

Related and worth deciding together: these are Thrill Studio skills and nothing says so. Measured on the published page for `apca-contrast`:

- skills.sh renders the **SKILL.md body**. Frontmatter is not surfaced — `kind`, `applies_to`, `review_mode`, `voice_scope`, `License`, `Author`, `Publisher` all return **0 hits** on the rendered HTML.
- Attribution today is the breadcrumb only: `skills / thrillmade / agent-skills / apca-contrast`. Nothing names Thrill Studio, nothing points anywhere.

So a `source:` frontmatter field would be free against the 8192 cap (`validate_skills.py:205` starts the body after the frontmatter) and **invisible to every human who reads the page**. Branding has to be in the body to exist at all.

One line, last before `## Sources`, carrying both the attribution and the link up to the directory. Gate it the same way — present exactly once in every skill, so it cannot drift or be trimmed away under budget pressure.

**Byte cost, stated honestly:** roughly 40 bytes x 49 skills, and a handful sit close enough to the cap that they need a trim first. That trim is now a visible, declared act rather than a silent one — which is what #218 exists for.

## Why this is worth doing rather than filing and forgetting

Every failure this catalog has had recently is the same shape: a fact with more than one owner, and no gate keeping the copies honest. The page-type count in `arlyn-working` has four owners and three disagree with the code. #217 is the catalog and a subscriber diverging in both directions with nothing recording it. The README directory is one more copy of the same fact, currently true, with nothing holding it there.

A generated, gated, travelling directory is the version of this that cannot rot.

thrillmot:
Closes #229 (the half that can be closed — see the end).

An agent in a subscriber repo can list 29 directory names and learn nothing about what any of them do. The failure isn't slow reading, it's **concluding coverage is absent and writing the guidance again** — which the arlyn lane hit for real: *"the studio-skills survey found twelve rules I was about to write that already existed, caught only because it read all 49 skills in full. No ordinary task does that."*

This ships the map **as a skill**, so it travels with `npx skills add` rather than staying behind in the README.

## Generated, not hand-kept

#229's first finding is that the README table is *"complete by diligence, and nothing keeps it that way — the next skill added is the one that breaks it, silently."* A hand-kept skill body would recreate that bug one level down. So:

- `gen_skill_directory.py` renders the body from the gated source
- `validate_skills.py` fails if the committed body differs from what the generator produces
- **Proved on my own mutation, not on report:** appending one HTML comment to the generated body → `--check` **exit 1**; restore → **exit 0**

## It closed a real defect it wasn't looking for

`skillforge` is **deprecated** and the catalog was routing to it as live. Rather than fix the row, the lane closed the class: `validate_skills` now fails when a skill announcing `SUPERSEDED` (via description, `superseded_by`, or `status`) sits outside the `deprecated` family. Break/restore verified in-tree with the mutation grepped to confirm it landed.

## Three refusals worth reading

- **It deleted its own growth table.** The build's projected ceilings (74/81/87/155) did not reproduce. Replaced with `levers()`, which *measures* on demand — today: `50 skills, this format holds 68; owns capped at 24 bytes, 70; family lines halved, 77; bare names, 124`. A test pins that the size-refusal message recites no measured number.
- **It rejected generating the README table from `owns`.** A 32-byte fragment is worse prose than the sentences already there — that makes the README worse in order to make it derived.
- **It rejected the per-skill footer link**, and the arithmetic matches #238's independently: 8 skills have <65 bytes of headroom, `test-discipline` has **1**. Each footer needs a `docs/prose-removals.md` row. **That is #229's other half and stays open.**

## The self-catch worth more than the fixes

One of its own 37 mutations was **inert and it rewrote the code rather than keep the green**. `re.compile(r"^\s*SUPERSEDED\b")` used with `.match()` has a redundant `^` — removing either anchor leaves the other anchoring, so neither is testable and the mutation stayed green. The regex is now `r"SUPERSEDED\b"` with `.match()` as the single owner of anchoring, and the mutation (`.match` → `.search`) goes red. `\s*` went too: YAML strips leading whitespace in every scalar style this catalog uses, so it had no reachable case.

That is the exact failure this repo keeps finding — a guard that cannot fail — caught by the lane against itself.

## Gates, exit codes read from the process

```
validate_skills.py                    OK: 50 skills validated cleanly.              EXIT=0
gen_skill_directory.py --check        OK: … is current (6431 body bytes,
                                      1469 under the reserve; room for 18 more)     EXIT=0
check_prose_retention.py              OK: 1 changed SKILL.md. No undeclared removal. EXIT=0
logmind check-links                   All markdown links resolve                    EXIT=0
pytest tests/ -q                      704 passed in 332.43s                         EXIT=0
mutation suite                        39 passed (37 mutations + control + registry) EXIT=0
```

Body **6430 bytes**, headroom **1762** against the cap and **1469** against the reserve.

## Reviewer notes

- **`check-links` reporting green here means nothing** for the skill bodies (#234) — it never walks `skills/`. Links were verified by hand.
- Two honest limits recorded in the generator docstring rather than papered over: `dev` is not installable (only `main` is reachable by `npx skills add`), and 6 of 8 declared subscribers do not use the CLI at all.
- The description leads with *"the thrillmade/agent-skills catalog — that catalog only, not the wider skills.sh ecosystem"* after a routing collision with `find-skills` was found on a cheap model.


thrillmot:
## Verdict: MERGE

PR #243 (`feat/catalog-directory` → `dev`, head `ec4e441`) adds a generated catalog-directory skill (`skills/finding-a-catalog-skill`) plus the gate that keeps it honest. I worked read-only from an archive of `origin/feat/catalog-directory` and a second scratch copy for destructive drift testing; nothing was written under the real repo tree (`git status --porcelain` there shows no changes from this review). This is the independent adversarial review the repo's own rule requires for a `dev` merge — `clud-bug-review` is confirmed `neutral`/"review unavailable" on this PR's actual head SHA (re-checked live, not assumed from the boilerplate), so nothing else has reviewed this.

### 1. Can the directory go stale while every check stays green? No — verified live, not reasoned about.

I ran all five drift vectors named in the brief against a live checkout and read the gate's actual output each time:

| Drift | Result |
|---|---|
| Add a skill, don't regenerate | **3 errors**: missing map entry, unrenderable directory, missing README row |
| Regenerate the map entry but not the directory body | **1 error**, isolated exactly to the byte-comparison: *"this body is GENERATED and no longer matches..."* |
| Hand-edit the generated body directly | Same staleness error, fires immediately |
| Rename a skill dir without updating the map | **5 errors** across map, directory, and README (both directions) |
| Delete a skill whose map+README+directory entries still reference it | **3 errors** |
| Delete the generator script itself | Gate fails closed with a named `::error::` annotation (not a bare traceback) — confirmed the `try/except ImportError` guard around `import gen_skill_directory` actually does what its comment claims |

One real (documented, tested) seam: deleting the directory skill's dir **and** its placement-map entry **and** every README link to it in the same change goes green in `validate_skills.py` alone — I reproduced this live (49 skills validated cleanly). This is not a hidden hole: `directory_errors()`'s own docstring names it, and `tests/test_gen_skill_directory.py::test_this_catalog_publishes_a_directory` exists specifically to close it — I confirmed that test fails correctly in exactly this scenario. Filed as nonblocking below since the mitigation is real and tested, but it does mean the invariant currently has exactly one owner.

### 2. The SUPERSEDED rule: fires on real retirement, silent on the word in passing — confirmed both directions, live.

Rather than trust the committed unit tests (which are thorough — `test_the_word_superseded_mid_sentence_is_not_a_retirement`, `test_the_superseded_marker_is_read_at_the_start_of_the_description`, `test_the_reserved_retirement_keys_count_too`), I reproduced both edges against the real gate on a live tree:

- A skill whose description reads *"Judging whether a rule was SUPERSEDED by another one, without being itself superseded"*, filed under a live family → **0 errors**. No false positive.
- A skill whose description opens *"SUPERSEDED by ..."*, filed under a live (non-`deprecated`) family → **fires**, with the exact message naming the fix (`File it under 'deprecated'`).

I also grepped the real 50-skill catalog for every case-insensitive "supersed" occurrence: `curating-a-skill-catalog`'s body mentions "superseded API" in prose (never scanned — the rule only reads frontmatter `description`/`superseded_by`/`status`, never the body), and the three actually-deprecated skills (`skillforge`, `component-sizing`, `design-token-naming`) all open their description with `SUPERSEDED` and are all filed under `deprecated`. The real tree validates clean (`OK: 50 skills validated cleanly`), which is itself a control: if the rule over-fired anywhere in the live catalog, this would be red.

### 3. The 37 committed mutations: all real, verified past the harness's own report.

`grep`-counted 37 distinct mutation names in `tests/test_skill_directory_mutations.py` (matches the claim exactly). The full suite passed with a properly-captured, unpiped exit code: **704 passed, exit 0** — I initially made the exact `cmd | tail; echo $?` mistake the brief warned about and caught it myself; the redone, direct capture confirms real exit 0.

Beyond trusting the harness, I hand-verified 3 mutations outside it: applied the exact text replacement to a fresh checkout, grepped the mutated line to confirm it landed at the claimed spot, then ran the real test file directly and read the actual failure:

- `the_owns_cap_is_off_by_one` — landed at line 552; the dedicated boundary test (`test_owns_one_byte_over_the_cap_is_rejected`) failed for exactly the right reason (no error where one was expected).
- `the_size_reserve_is_off_by_one` — landed at line 553; 3 tests failed, including a `TypeError` when `size_error()` returned `None` past the shifted boundary — a real, targeted failure, not noise.
- `an_unhashable_family_crashes_the_gate` — landed at line 582; reproduced the *exact* historical regression named in the code comment (`family: []` in JSON → `TypeError: unhashable type: 'list'` crashing the whole gate instead of being reported by it). This is the one the comment says "shipped, briefly."

On the reported "one mutation was inert" history: that was caught and fixed *within* the first commit (405b6c1, before this PR's current state) — the commit message names it (`[] or [x] is [x]`). I didn't find a second one: all 37 mutations pass their own `source.count(old) == 1` and `mutated != source` landing-assertions in the real (unpiped) full-suite run, and the 3 I checked by hand each produced a real, on-target failure rather than an incidental one.

### 4. Growth ceiling: re-derived independently, matches exactly.

Ran the generator directly against the real checked-out tree (not the test suite, not the commit message):

```
$ python3 .github/scripts/gen_skill_directory.py
OK: skills/finding-a-catalog-skill/SKILL.md is current (6431 body bytes, 1469 under the reserve; room for 18 more skills at today's density).
$ python3 -c "...ceiling(...)"
dirs: 50, ceiling: 68, room: 18
```

This matches the decision log's *second* entry (post-panel-refutation: "measured: 6431 body bytes, room for 18 more skills") exactly, not the *first* commit's now-superseded 74/79/155 table — the PR is honest in its own decision log that the first table "reproduced under none of the models it stated" and was deleted in favor of `ceiling()`, a function that renders the real tree with synthetic skills at today's measured density rather than reciting a number. I independently reproduced the number it now measures; there is nothing hand-typed left to disagree with the tree.

### 5. Gates, run for real:

- `python3 .github/scripts/validate_skills.py` → `OK: 50 skills validated cleanly.` (exit 0)
- `python3 .github/scripts/gen_skill_directory.py --check` → current, matches the report above (exit 0)
- `check_prose_retention.py --base <merge-base> --head origin/feat/catalog-directory` (run directly against the real repo's git history, read-only) → `OK: 0 changed SKILL.md file(s), plus 1 added and 0 deleted whole... No undeclared prose removal.` (exit 0)
- Full suite: **704 passed**, exit 0, captured directly (not through a pipe)
- PR's actual GitHub checks (`gh pr checks 243`): `check prose retention`, `check-decisions`, `check-derived-docs`, `check-links`, `test`, `validate skills` all `pass`; `clud-bug-review` confirmed `neutral`/unavailable on this exact head SHA via the check-runs API, matching AGENTS.md's documented state.
- Links inside `skills/finding-a-catalog-skill/SKILL.md` (the one SKILL.md this PR touches, and the one `check-links` is documented not to see): hand-checked all 3 external URLs (200 OK) and the one cross-branch tree link, confirmed the target path exists on `main` via `git cat-file -e`.

### Net

Every mechanism the brief asked me to distrust-and-verify held up under direct, live reproduction rather than a re-read of the report. The one real coverage seam (simultaneous deletion of the directory's dir + map entry + all README links) is disclosed in the code's own docstring, has a dedicated test that I confirmed catches it, and is listed above as nonblocking rather than something asking for a fix in this PR.

PR #245, opened 2026-08-18
The prose gate never compared added files — a skill could be added and gutted on the same branch, green

thrillmot:
The prose-retention gate did not compare files **added** within the base..head range. So an author could add a skill in one commit and gut it in the next commit on the same branch, and the gate reported success.

Found by a reviewer's own control while verifying #233 — not by the gate.

## The gap, reproduced both before and after

Exit codes captured into a variable, never read after a pipe:

```
BEFORE   add a skill (50 words), gut it in the next commit   EXIT=0   "…plus 1 added and 0 deleted
                                                                       whole, which this gate does
                                                                       not compare"
AFTER    identical case                                       EXIT=1   "this SKILL.md lost 53 words
                                                                       that nothing in the same part
                                                                       of it replaced"
CONTROL  added once, never touched again                      EXIT=0   silent — no false positive
```

**This was live on #233**, which adds five skills, so the gate certified nothing about their substance.

## The fix

For a path git reports as `A` between base and head, `_first_appearance()` walks `git log --reverse base..head -- path` and returns the earliest commit that touched it; `collect()` compares head against **that** rather than skipping the file. `Diff.pair`'s existing `before != after` guard means a file touched once produces no case — silent and correct, with no special-casing.

Cases pinned by test:

| case | behaviour |
|---|---|
| added, never modified again | silent |
| added, modified several times | compares against the **first** appearance, not the predecessor — the test constructs a case where comparing to the predecessor nets zero (pure rewrap) but the first-appearance comparison still finds a 10-word cut two commits back |
| added then deleted within the range | invisible to `git diff base head` in either tree, so **as invisible to this gate as to a human reading the Files view** — pinned as such rather than claimed OK |
| unwalkable history (shallow clone) | **hard error** `UNWALKABLE_ADDITION`, matching the existing base-resolution stance. Never report success from an absence. |

## Honesty preserved

The gate already *declared* this limitation, which is why it was findable at all. That honesty is kept: the OK line no longer lumps added files into "which this gate does not compare" — it now reads `"N added, each compared against its first version within this range"`, distinct from `"M deleted whole, which this gate does not compare"`. Deleted files remain genuinely out of scope and say so.

## Verification

**Red before green** — 7 tests written first, confirmed failing against unfixed source (`AttributeError: no attribute '_first_appearance'`, wrong exit codes, stale message text), then implemented.

**Three mutations added** to the existing harness, all confirmed red:
- `added_files_stop_being_compared_to_their_first_appearance` — restores the exact original gap
- `first_appearance_compares_against_the_last_touch_not_the_first` — `shas[0]` → `shas[-1]`
- `unwalkable_added_history_silently_treated_as_unchanged` — `return None` → `return head`

Plus a manual mutate/restore on the primary guard with the mutation **grepped in the tree** to confirm it landed before trusting the red.

**Full suite: 581 passed, exit 0** (captured via redirect and explicit `echo $?`).

## Against #233's real branch

```
OK: 6 changed SKILL.md file(s), plus 5 added, each compared against its first version
within this range. No undeclared prose removal.
```

**Not a false positive** — the 6 changed are the pre-existing skills that branch modifies (`git diff --name-status` confirms `M` on each); the 5 added are each a single commit, so comparing against their own first appearance correctly finds nothing, because there is genuinely no gutting commit there.

To prove the fix would actually fire on that branch rather than merely stay quiet, a detached follow-up commit was built with plumbing (`git commit-tree` — no checkout, no ref, nothing touched in any working tree) gutting `## Verification` out of `usability-heuristics`. The gate fired: `exit 1`, naming that file and that section, with no other file flagged.

### DIFF
--- a/.github/scripts/check_prose_retention.py
+++ b/.github/scripts/check_prose_retention.py
@@ -1403,12 +1403,14 @@ def hatch_state(before: str, after: str, skill: str, net: int) -> str:
 # The one phrase that identifies each failure mode's annotation, owned here so
 # the message and the guidance `main` prints under it cannot drift apart --
 # `run` also annotates the ledger itself and any file it could not scope, and
-# `main` annotates any blob that would not come back. Those are four different
-# findings and they do not share a remedy. LOST_PROSE additionally carries the
-# count of undeclared removals.
+# `main` annotates any blob that would not come back and any added file whose
+# own history could not be walked. Those are five different findings and they
+# do not share a remedy. LOST_PROSE additionally carries the count of
+# undeclared removals.
 LOST_PROSE = "::this SKILL.md lost "
 UNSCOPABLE_FILE = "::this SKILL.md's YAML frontmatter could not be located"
 UNREADABLE_BLOB = "::git lists this file as changed"
+UNWALKABLE_ADDITION = "::this SKILL.md was added within this range, but the commit"
 LEDGER_REWOUND = "::this change takes "
 SLOTS_WITHDRAWN = f"back OUT of {LEDGER}"
 
@@ -1440,13 +1442,19 @@ def run(
 ) -> list[str]:
     """Check every (path -> (before, after)) pair. Returns error annotations.
 
-    `cases` carries every SKILL.md modified between the two revisions, renames
-    included, keyed by its path at head. Whole-file additions and deletions are
-    deliberately out of scope: git renders those as a file appearing or
-    disappearing, which review cannot miss, and `main` says out loud how many
-    it skipped. This gate exists for the removal that hides inside an
-    otherwise-ordinary edit -- which is how all three historical cases got
-    through.
+    `cases` carries every SKILL.md this gate could compare between the two
+    revisions -- modified, renamed, and added-then-changed within the range,
+    keyed by its path at head. A file added and never touched again has
+    nothing in `cases` for it: compared against the commit that introduced
+    it, its own first version, the loss is zero by construction, and that
+    silence is correct rather than a gap this function has.
+
+    Whole-file DELETIONS alone stay deliberately out of scope: a dead file has
+    nothing left to protect, git renders the deletion as a file disappearing,
+    which review cannot miss, and `main` says out loud how many it skipped.
+    This gate exists for the removal that hides inside an otherwise-ordinary
+    edit or an add -- which is how all three historical cases, and the gap
+    that let a whole-file add through undeclared, got through.
     """
     ledger = LedgerDiff(ledger_before, ledger_after)
     errors: list[str] = []
@@ -1596,6 +1604,14 @@ REMEDIES = [
         False,
     ),
     (
+        UNWALKABLE_ADDITION,
+        "Each file named above was listed as added, and the commit that first "
+        "introduced it inside the range could not be found, so it was never "
+        "compared against that first version. Check the checkout step sets "
+        "`fetch-depth: 0`.",
+        False,
+    ),
+    (
         SLOTS_WITHDRAWN,
         f"Put back every line this change took out of {LEDGER}. While anything "
         f"is missing from that file, no row the change adds counts at all.",
@@ -1703,13 +1719,50 @@ def _show(rev: str, path: str) -> str | None:
         raise Undecodable(rev, path) from None
 
 
+def _first_appearance(base: str, head: str, path: str) -> str | None:
+    """The earliest commit in `base..head` that touched `path`.
+
+    Called only for a path git's own diff already calls ADDED between `base`
+    and `head` -- the file does not exist at `base`, so some commit inside the
+    range must have created it, and that commit's own version of the file is
+    what a whole-file add should be compared against instead of `base`, where
+    it is not there to compare against at all. `base` itself was never a
+    candidate: comparing to the version at `base` is exactly what a plain
+    modification does, and this path does not have one.
+
+    None only when the range could not be walked at all -- a clone too
+    shallow to enumerate `base..head` for this path. That is the same
+    shallow-clone failure `main`'s own base-resolution guard already refuses
+    on rather than silently passing over, so a caller gets `None` back rather
+    than an empty result it could mistake for "no commits touched this path",
+    which cannot happen for a path git itself calls added.
+    """
+    got = _git("log", "--reverse", "--format=%H", f"{base}..{head}", "--", path)
+    if got.returncode != 0:
+        return None
+    shas = [ln for ln in got.stdout.splitlines() if ln]
+    return shas[0] if shas else None
+
+
 class Diff:
     """What changed between two revisions, split into what this gate can judge.
 
-    `cases` are the modified and renamed SKILL.md files, keyed by their path at
-    head. `added` and `deleted` are counted but not compared, and `main` says
-    so out loud: a gate that prints OK without mentioning what it left out is
-    reporting success over a comparison it did not make.
+    `cases` are the modified, renamed and added-then-changed SKILL.md files,
+    keyed by their path at head. A file added within the range is compared
+    against the commit that first introduced it, not against `base`, where it
+    does not exist -- so a skill added in one commit and gutted in a later one
+    on the same branch is still caught. A file added and never touched again
+    compares against itself and has nothing in `cases` for it, which is
+    correct: the loss is zero by construction, not a comparison skipped.
+
+    `deleted` is counted but not compared -- a dead file has nothing left to
+    protect. `unresolved_added` is the one whole-file-add case still out of
+    scope on purpose: a path git calls added whose own history inside
+    `base..head` could not be walked, the shallow-clone failure `main`'s own
+    base-resolution guard already refuses on rather than passes over. `main`
+    says so out loud for both `deleted` and `unresolved_added`: a gate that
+    prints OK without mentioning what it left out is reporting success over a
+    comparison it did not make.
     """
 
     def __init__(self) -> None:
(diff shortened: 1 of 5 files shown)


thrillmot:
## Verdict: MERGE

PR #245 closes a real gap and I reproduced the gap, the fix, and every one of the five things this review was asked to check — independently, in throwaway repos, not by reading the PR's own tests and trusting them.

### 1. The gap and the fix — CONFIRMED
Built alpha (base) + gamma (added in full, one commit) + gamma gutted (very next commit, same branch). Against `origin/dev`'s script:
```
OK: 0 changed SKILL.md file(s), plus 1 added and 0 deleted whole, which this gate does not compare. No undeclared prose removal.
EXIT:0
```
Against the fix branch's script, same history:
```
::error file=skills/gamma/SKILL.md::this SKILL.md lost 85 words that nothing in the same part of it replaced...
EXIT:1
```
Control (added once, never touched again — must stay silent):
```
OK: 0 changed SKILL.md file(s), plus 1 added, each compared against its first version within this range. No undeclared prose removal.
EXIT:0
```
Both directions hold: the gap existed, the fix closes it, and the fix does not fire on ordinary skill creation.

### 2. The unwalkable-history path — CONFIRMED, with a stronger repro than the PR's own test
The PR's own test forces `_first_appearance` to return `None` via `monkeypatch`. I wanted a genuine failure instead. A plain shallow clone doesn't reach the new code path — it trips `main`'s *pre-existing* base-resolution guard first, because `--depth 1` only fetches the head commit, so `base` itself never resolves. I confirmed that too, then went further: I corrupted the intermediate "add gamma" commit object on disk directly, leaving `base` and `head` independently resolvable via `git rev-parse --verify` (both succeed). That forces `_first_appearance`'s own `git log base..head -- path` call to fail with a nonzero return code — the exact `got.returncode != 0` branch. Result: exit 1, correct message, correct `fetch-depth: 0` remedy, never a silent OK.

### 3. Added-then-deleted within the range — CONFIRMED, not just asserted
Built the history (add gamma, delete gamma, same branch). `git diff --name-status base head` produces nothing for gamma — genuinely absent from both trees, same as a human reading the PR's Files view would see. The gate's own output: `OK: 0 changed SKILL.md file(s). No undeclared prose removal.` — zero mention of gamma anywhere. The PR's claim that this case is invisible to the gate exactly as it is to review is real, not rationalized.

### 4. The honesty change on the OK line — CONFIRMED with both cases present together
Built a history with one added and one deleted skill in the *same* commit (not tested in isolation, which could hide a wording collision). Output:
```
OK: 0 changed SKILL.md file(s), plus 1 added, each compared against its first version within this range; 1 deleted whole, which this gate does not compare. No undeclared prose removal.
```
Added files get their own honest clause; deleted files keep the old "which this gate does not compare" wording. Neither is dropped when both are present.

### 5. Mutation suite, full suite, and the real PR #233 branch
- `tests/test_check_prose_retention.py` + `tests/test_prose_retention_mutations.py` run from a clean `git archive` of the branch (no working-tree contamination): **466 passed**.
- The 3 new mutation-guard entries isolated by name and run alone: **3 passed** — each of `added_files_stop_being_compared_to_their_first_appearance`, `first_appearance_compares_against_the_last_touch_not_the_first`, and `unwalkable_added_history_silently_treated_as_unchanged` is genuinely caught by the suite, not just present in the list.
- `validate_skills.py`: `OK: 51 skills validated cleanly.`
- Real branch check (task's explicit ask): fetched PR #233's live branch (`feat/nominate-thrill-studio-skills`, 5 added skills), computed `merge-base(dev, branch)` the same way CI does. Pre-fix script: `OK: 6 changed..., plus 5 added...which this gate does not compare` exit 0 — reproduces the exact false-negative the decision doc cites. Post-fix script: `OK: 8 changed..., plus 5 added, each compared against its first version within this range` exit 0, no undeclared removal — **no false positive against a real, unrelated, currently-open branch.**
- Full local `pytest tests/ -q` did not finish — see "could not check" for why and what I substituted (CI's own green `test` check on this exact head commit, a843e6e, 2m25s).

### Scope and hygiene
`git diff --name-only origin/dev...origin/fix/prose-gate-added-files` is exactly 5 files: the script, its two test files, the branch decision doc, and `docs/timeline.md`. No workflow file touched (the CI-invocation contract in `.github/workflows/check-prose-retention.yml` is untouched). Decision doc and timeline entries are present and consistent with the diff.

### Non-blocking observation (not one of the five checklist items, not verified either way)
`_first_appearance`'s inner `git log --reverse -- path` has no `--follow`. For a path that reaches its final name through an in-range rename *and* is separately classified by git's outer diff as a pure add (a rename below the 25% similarity floor, which `collect()`'s own docstring already discusses for the delete half), the "first appearance" found could in principle be the rename commit rather than the file's true origin elsewhere. I did not construct this case — it's a real gap in coverage, not a bug I confirmed, and it sits outside this PR's stated scope. Worth a follow-up test, not a blocker here.

**No blocking findings.** clud-bug-review on this PR is `neutral` ("review unavailable"), consistent with AGENTS.md's documented outage since #188 — this panel is the review the repo's rule requires, and it's independent (I did not write this fix).

thrillmot:
## Verdict: MERGE

PR #245 closes a real gap and I reproduced the gap, the fix, and every one of the five things this review was asked to check — independently, in throwaway repos, not by reading the PR's own tests and trusting them.

### 1. The gap and the fix — CONFIRMED
Built alpha (base) + gamma (added in full, one commit) + gamma gutted (very next commit, same branch). Against `origin/dev`'s script:
```
OK: 0 changed SKILL.md file(s), plus 1 added and 0 deleted whole, which this gate does not compare. No undeclared prose removal.
EXIT:0
```
Against the fix branch's script, same history:
```
::error file=skills/gamma/SKILL.md::this SKILL.md lost 85 words that nothing in the same part of it replaced...
EXIT:1
```
Control (added once, never touched again — must stay silent):
```
OK: 0 changed SKILL.md file(s), plus 1 added, each compared against its first version within this range. No undeclared prose removal.
EXIT:0
```
Both directions hold: the gap existed, the fix closes it, and the fix does not fire on ordinary skill creation.

### 2. The unwalkable-history path — CONFIRMED, with a stronger repro than the PR's own test
The PR's own test forces `_first_appearance` to return `None` via `monkeypatch`. I wanted a genuine failure instead. A plain shallow clone doesn't reach the new code path — it trips `main`'s *pre-existing* base-resolution guard first, because `--depth 1` only fetches the head commit, so `base` itself never resolves. I confirmed that too, then went further: I corrupted the intermediate "add gamma" commit object on disk directly, leaving `base` and `head` independently resolvable via `git rev-parse --verify` (both succeed). That forces `_first_appearance`'s own `git log base..head -- path` call to fail with a nonzero return code — the exact `got.returncode != 0` branch. Result: exit 1, correct message, correct `fetch-depth: 0` remedy, never a silent OK.

### 3. Added-then-deleted within the range — CONFIRMED, not just asserted
Built the history (add gamma, delete gamma, same branch). `git diff --name-status base head` produces nothing for gamma — genuinely absent from both trees, same as a human reading the PR's Files view would see. The gate's own output: `OK: 0 changed SKILL.md file(s). No undeclared prose removal.` — zero mention of gamma anywhere. The PR's claim that this case is invisible to the gate exactly as it is to review is real, not rationalized.

### 4. The honesty change on the OK line — CONFIRMED with both cases present together
Built a history with one added and one deleted skill in the *same* commit (not tested in isolation, which could hide a wording collision). Output:
```
OK: 0 changed SKILL.md file(s), plus 1 added, each compared against its first version within this range; 1 deleted whole, which this gate does not compare. No undeclared prose removal.
```
Added files get their own honest clause; deleted files keep the old "which this gate does not compare" wording. Neither is dropped when both are present.

### 5. Mutation suite, full suite, and the real PR #233 branch
- `tests/test_check_prose_retention.py` + `tests/test_prose_retention_mutations.py` run from a clean `git archive` of the branch (no working-tree contamination): **466 passed**.
- The 3 new mutation-guard entries isolated by name and run alone: **3 passed** — each of `added_files_stop_being_compared_to_their_first_appearance`, `first_appearance_compares_against_the_last_touch_not_the_first`, and `unwalkable_added_history_silently_treated_as_unchanged` is genuinely caught by the suite, not just present in the list.
- `validate_skills.py`: `OK: 51 skills validated cleanly.`
- Real branch check (task's explicit ask): fetched PR #233's live branch (`feat/nominate-thrill-studio-skills`, 5 added skills), computed `merge-base(dev, branch)` the same way CI does. Pre-fix script: `OK: 6 changed..., plus 5 added...which this gate does not compare` exit 0 — reproduces the exact false-negative the decision doc cites. Post-fix script: `OK: 8 changed..., plus 5 added, each compared against its first version within this range` exit 0, no undeclared removal — **no false positive against a real, unrelated, currently-open branch.**
- Full local `pytest tests/ -q` did not finish — see "could not check" for why and what I substituted (CI's own green `test` check on this exact head commit, a843e6e, 2m25s).

### Scope and hygiene
`git diff --name-only origin/dev...origin/fix/prose-gate-added-files` is exactly 5 files: the script, its two test files, the branch decision doc, and `docs/timeline.md`. No workflow file touched (the CI-invocation contract in `.github/workflows/check-prose-retention.yml` is untouched). Decision doc and timeline entries are present and consistent with the diff.

### Non-blocking observation (not one of the five checklist items, not verified either way)
`_first_appearance`'s inner `git log --reverse -- path` has no `--follow`. For a path that reaches its final name through an in-range rename *and* is separately classified by git's outer diff as a pure add (a rename below the 25% similarity floor, which `collect()`'s own docstring already discusses for the delete half), the "first appearance" found could in principle be the rename commit rather than the file's true origin elsewhere. I did not construct this case — it's a real gap in coverage, not a bug I confirmed, and it sits outside this PR's stated scope. Worth a follow-up test, not a blocker here.

**No blocking findings.** clud-bug-review on this PR is `neutral` ("review unavailable"), consistent with AGENTS.md's documented outage since #188 — this panel is the review the repo's rule requires, and it's independent (I did not write this fix).

PR #249, opened 2026-08-18
The census files a digest every week and closes none — close the prior one on each run

thrillmot:
CEO-approved. The weekly census files a digest issue every run and closes nothing:

```
grep 'close|supersede|previous' in skill-census.yml  ->  0 occurrences
CONTROL 'gh issue' in the same file                  ->  8
```

Four digests are open and only the newest is current. A weekly snapshot is superseded **by construction** the moment the next one files — so this was an issue tracker nobody prunes, which trains everyone to ignore it, which is what makes the real issues invisible.

## The hard part was not closing — it was not closing the wrong thing

The census also files `[gap]` / `[placement]` / `[revise]` verdict issues, and **those are real work that must stay open.** The guard requires **both** signals together:

```jq
[.[] | select((.labels | length) == 1)
      | select(.title | test("^census [0-9]{4}-W[0-9]{2} digest$"))
      | select(.title != $t)]
```

- `labels length == 1` — every verdict carries a second label on top of `census`; a digest carries only `census`. This rejects a verdict whose declared labels all fell through the allow-list.
- **anchored** title regex — the deterministic string the workflow itself generates, so a human-authored issue like *"the census digest generator double-counts untracked skills"* survives.
- `.title != $t` — excludes the cycle's own digest, which is what makes a re-run idempotent.

**Neither signal alone is safe**, and that was proven rather than argued — see mutation 2.

Checked against the **live backlog** (`gh issue list --label census`), not inferred from the YAML: every digest ever filed matches, every verdict is correctly excluded, including the four already-open stale digests this exists to clean up.

## Mutation testing — and one mutation that didn't fire

Three clauses, each broken → grepped in the tree → confirmed red → restored → confirmed green.

**Mutation 2 initially stayed green.** Loosening the anchored regex to `test("digest")` changed nothing, because the fixture only had a case the *other* clause already rejected — so the mutation wasn't isolating the clause it aimed at. Rather than accept that, the lane added a fixture case specifically to probe it independently, and then the mutation fired. **An inert mutation caught and fixed, not banked.**

The other two: loosening `length == 1` to `>= 1` closed a verdict-shaped issue; dropping the idempotency exclusion made a re-run close the digest it had just filed.

Final diff verified byte-identical to the saved fix after all three mutate/restore cycles.

## Deliberately not done

- **No in-body marker.** Label+title already identifies the real backlog uniquely, including the four digests filed *before* this fix existed — a body marker would only help issues filed afterwards and would do nothing for the ones this exists to clean up.
- **The four stale digests are not closed by hand.** That is a live mutation outside this brief's scope; the workflow closes all but the newest on its next real run.
- **Digests keep being filed.** The per-cycle record is what a liveness check reads to notice the census has *stopped* — a census that silently stops looks identical to one with nothing to report, and that gap is already named in `skdd`'s STATUS.md as the load-bearing half nobody has built.

## Gates

```
yaml.safe_load          parses; 9 steps, same id sequence
bash -n                 file_issues run-script OK
validate_skills.py      OK: 56 skills validated cleanly    EXIT=0
pytest tests/ -q        1035 passed  (1030 baseline + 5)   EXIT=0
```

### DIFF
--- a/.github/workflows/skill-census.yml
+++ b/.github/workflows/skill-census.yml
@@ -21,9 +21,10 @@ name: skill-census
 # secret-guard step fails loudly with remediation if they are not.
 #
 # Idempotency: one digest issue per ISO-week cycle (commented on re-run, not
-# duplicated); verdict issues de-duplicated by exact title; hard cap of 5
-# verdict issues filed per cycle with the overflow count surfaced in the
-# digest. See the "File issues" step.
+# duplicated); every OTHER open digest is closed as superseded once this
+# cycle's is filed, so exactly one stays open at a time; verdict issues
+# de-duplicated by exact title; hard cap of 5 verdict issues filed per cycle
+# with the overflow count surfaced in the digest. See the "File issues" step.
 
 on:
   schedule:
@@ -241,6 +242,7 @@ jobs:
             --json number,title 2>/dev/null \
             | jq -r --arg t "$DIGEST_TITLE" 'map(select(.title == $t)) | .[0].number // empty') || EXISTING_DIGEST=""
 
+          DIGEST_NUM="$EXISTING_DIGEST"
           if [ "$DRY_RUN" = "true" ]; then
             echo "=== DRY-RUN: digest issue ==="
             if [ -n "$EXISTING_DIGEST" ]; then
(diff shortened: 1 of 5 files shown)


thrillmot:
## Panel: MERGE — verified against the live backlog, not just fixtures

The panel ran the exact jq guard against the **real** `gh issue list --label census` output (read-only) rather than only the test fixtures:

```
next-cycle simulation  -> 228 190 184 177   (exactly the four stale digests)
idempotency (W34)      -> 190 184 177        (correctly drops today's own digest)
every open verdict     -> excluded           (#191, #186, #185, #149…#137)
```

Then 10 adversarial synthetic issues — a verdict whose extra labels all fell through the allow-list, a human-authored look-alike title, a manually `census`-labelled issue, lower-case `w`, trailing whitespace, 1- and 3-digit week numbers — **zero false positives.**

It also **re-fired the mutation that initially stayed green**, rather than taking that on report: loosening the anchored regex to `test("digest")` now fails with `issue #997 was closed but must not be` — the human-authored look-alike. The clause genuinely tests something.

Dry-run was tested with a scripted mock `gh` first on `PATH`: `DRY_RUN=true` printed three close lines and the mock's call log showed **zero** mutation calls; `DRY_RUN=false` on the identical fixture closed exactly #177/#184/#190 and left #228 alone.

### One real finding, non-blocking, filed rather than folded in

If `gh issue list` fails, `|| PRIOR_DIGESTS=""` swallows it: no annotation, close silently skipped, step exits 0. That is this repo's recurring defect shape — but three things make it a follow-up rather than a blocker, and the panel established all three:

1. **It fails safe, not dangerous** — it can only skip a close, never cause a wrong one.
2. **It self-heals** — the guard has no same-cycle requirement, so any later successful run closes the same stale digest.
3. **It is not introduced here** — the identical `2>/dev/null | jq … || VAR=""` idiom already exists on `dev` in the `EXISTING_DIGEST` lookup. This PR follows the file's convention rather than adding a new hole.

Filed as a follow-up.

**Merging.** Gates re-derived independently: 1035 passed at head vs 1030 collected at `dev` — the +5 delta is exactly the new test file, not padding elsewhere.

Issue #225, opened 2026-08-15
Norman's forcing functions, constraints, mapping and the two gulfs are in zero skills — a consumer hit the gap from inside


thrillmot:
Norman's *Design of Everyday Things* chapter 4 material — **forcing functions, the constraint classes, natural mapping, and the gulfs of execution and evaluation** — is absent from the catalog, and a consumer just hit the gap from the inside.

## Measured

```bash
$ for t in "forcing function" "natural mapping" "gulf of execution" "gulf of evaluation" "affordance"; do
    printf "%-20s %s\n" "$t" "$(grep -rli "$t" skills/*/SKILL.md | wc -l)"; done
forcing function      0
natural mapping       0
gulf of execution     0
gulf of evaluation    0
affordance            2      # designing-elite-ui, visual-polish — mentioned, not owned
```

**Control:** the same grep for `contrast` returns 17, so the probe finds terms that are present. These zeros are real.

(`constraint` matches 7 skills, but on the ordinary English word — spacing constraints, agent constraints — not Norman's constraint *classes*. Not coverage.)

## How it surfaced

The arlyn lane researched `empirical-design-principles` to a **12,733-byte** draft and had to cut it to clear the 8,192 body cap. Chapter 4 is what went.

The routing consequence is the interesting part, and it is #200's defect arriving from a new direction: `empirical-design-principles` names `forcing function` in its description and **disclaims** it — handing the reader to its sibling `usability-heuristics`, where the material is now also absent. A disclaimer that points at an empty file is worse than no disclaimer, because it reads as a route.

They handled it correctly rather than papering over it: the sibling now states these topics are covered by **neither** skill and names the editions to cite. **An honest dead end beats a pointer to an empty file** — that framing is theirs and it is right.

## Why this is a legitimate split, not a size dodge

Our own size-gate error text (`validate_skills.py:224`) permits exactly one reason to split: *"split ONLY if this is genuinely two topics, never to hit the number."*

This clears that bar on evidence rather than assertion. Four of the topic's load-bearing concepts have **zero** occurrences across all 49 skills. It is not a slice of an existing skill that grew too big; it is a subject the catalog has never covered.

## Scope, and what this issue is NOT

**This must not land before the 20 Aug handoff, and it is not the arlyn lane's to carry.** They surfaced a real gap with a named cause while shipping something else; the declared dead end is the correct interim state and should stay exactly as they wrote it.

Candidate shape, for whoever picks it up:

[...]

PR #251, opened 2026-08-18
Guidance that was actively wrong: SC 2.5.8's missing exceptions, Norman's two gulfs, sRGB-default colour math (#235, #225, #170)

thrillmot:
Three content fixes. Unlike the last few batches these are not machinery — this is **guidance that was wrong in skills already shipping**, and one of them reaches a client on 20 Aug.

## #235 — the one with the date

`frontend-a11y` named **one** of SC 2.5.8's five exceptions. The missing **Inline** exception exempts targets "constrained by the line-height of non-target text" — which is every inline link in body copy. So the lens reported ordinary body text as an AA failure. For a rendered-surface lens, that is most pages, not an edge case.

Verified against the W3C text directly, not the issue's paraphrase — `curl` on the Recommendation and on Understanding 2.5.8. Both claims confirmed:

- five exceptions are **Spacing, Equivalent, Inline, User Agent Control, Essential**
- the Spacing text reads *"the circles do not intersect **another target**"* — never "another target's box". Bounding box appears only to locate the circle's **centre** for non-rectangular targets.

Fixed both. The box→target correction **returns 6 bytes**.

```
Inline|Equivalent|Essential|User Agent Control   0 -> 4
"another target's box"                           1 -> 0
CONTROL "2.5.8"                                  3    (probe still discriminates)
```

## #225 — the two gulfs, and why not in the obvious place

Mostly closed by #233 already; re-measured before planning. Forcing functions and mapping are now covered — **both gulfs were still zero**.

The lane rejected two plausible homes with reasons rather than picking the easy one:
- **`empirical-design-principles`** scopes itself to falsifiable *predictions*, and its own text declared the gulfs out of scope. Covering them there would fight the file's stated boundary.
- **`usability-heuristics`** is the better conceptual fit — same DOET ch.1 material — but has **12 bytes** of headroom. Landing it there would have meant a trim big enough to be the #213 defect.

Landed in `composing-a-screen` as a sixth Verification item, phrased as two checks: *execution — next action findable? evaluation — result visible?*

House rule honoured: **cite the chapter, not the count** (DOET carries both a six- and a seven-item list in different chapters, which has bitten this catalog before).

## #170 — colour

Two skills still gated on *"both models must pass"*, which forces a P3-native palette to satisfy an sRGB-era threshold it was never designed for. Matched to `oklch-color-space`'s existing wording rather than inventing a fifth phrasing.

## The retention gate refused this batch, correctly

Two removals, now declared with reasons:

| skill | words | why |
|---|---|---|
| empirical-design-principles | 6 | its disclaimer said the gulfs were "covered nowhere yet" — #225 covers them, so the sentence had become false about its own catalog |
| finding-a-catalog-skill | 12 | **generated** — the directory's "not here" entry for the gulfs, dropped by `gen_skill_directory.py` because the gap was filled |

The second is the interesting one: **a generated file can lose prose legitimately**, and the ledger is still the right place to say so, because a reader diffing the directory should not have to reverse-engineer why a line vanished.

*(Noting a trap for the next person: the gate reads **revisions**, not the working tree. Adding ledger rows without committing them leaves it failing, and it briefly looked like the rows had not worked.)*

## Headroom — tight, and disclosed

| file | before | after |
|---|---|---|
| frontend-a11y | 63 | **24** |
| composing-a-screen | 94 | **6** |
| empirical-design-principles | 62 | 61 |

Both lanes trimmed real duplication to fit and reported exactly what and why rather than trimming silently. One item was **explicitly deferred** for want of bytes — SC 2.5.8's "24 px" vs "24 px *diameter*" ambiguity, which the issue itself calls cosmetic.

## Gates

```
validate_skills.py         link scope: 370 relative / 51 absolute
                           OK: 56 skills validated cleanly              EXIT=0
check_prose_retention.py   OK: 7 changed SKILL.md file(s).              EXIT=0
```

### DIFF
--- a/.github/scripts/gen_skill_directory.py
+++ b/.github/scripts/gen_skill_directory.py
@@ -194,15 +194,13 @@ NOT_HERE_PROBES = [
         "the `superpowers` plugin, not here",
     ),
     ("test-driven", "**The same discipline under its other name**"),
-    (
-        # Narrowed 2026-08-18: `empirical-design-principles` (#233) filled the
-        # forcing-functions and mapping half of this gap, so the old probe term
-        # started matching and the generator correctly refused to publish a
-        # stale claim. The gulfs are still uncovered -- measured 0 files, with
-        # `contrast` -> 20 as the control that the matcher works. See #225.
-        "gulf of execution",
-        "**Norman's two gulfs** -- execution and evaluation",
-    ),
+    # Closed 2026-08-18 (#225): `composing-a-screen` now names both gulfs in its
+    # trigger surface and carries a checkable step for each (DOET rev. ed. 2013
+    # ch. 1), so `gulf of execution` stopped matching 0 and the probe was
+    # removed rather than narrowed -- nothing in this pair is still uncovered.
+    # Norman's constraints, forcing functions and natural mapping remain a real
+    # gap (declared in `empirical-design-principles`, not probed here because
+    # that skill's own disclaimer text would make the matcher self-trigger).
 ]
 
 # The control. A probe set that matches nothing proves nothing unless
--- a/docs/decisions-branches/fix__color-srgb-default.md
+++ b/docs/decisions-branches/fix__color-srgb-default.md
@@ -1,0 +1,13 @@
+← back to [docs/timeline.md](../timeline.md)
+
+## 2026-08-18 12:30 - Retire the 'both models must pass' contrast gate; finish chroma-harmonization's recompose fix (#170)
+
+**Reasoning:** APCA is UDTS's required perceptual gate; WCAG 2.2 is an optional, off-by-default cross-check, not a co-equal required model — apca-contrast and wcag-contrast still taught the pre-UDTS both-required framing after #176 fixed only oklch-color-space and chroma-harmonization's headline. apca-contrast also had no P3-meter rule at all, and chroma-harmonization was missing its two ADD items (recompose-at-shared-chroma, and the bottleneck-residual caveat).
+
+**Alternatives considered:** Could have left 'both must pass' as a stricter-than-required safety net; rejected because it forces P3-native palettes to satisfy an sRGB-era threshold never designed for them, exactly the defect the live UDTS incident was filed against.
+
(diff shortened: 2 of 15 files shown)


thrillmot:
## Panel: MERGE WITH FIX → both blockers fixed → merging

**Priority 1 verified clean by the panel against raw W3C**, not the PR's paraphrase: all five SC 2.5.8 exceptions correctly named, the Inline description a faithful paraphrase of the normative text, the box→target correction supported, and **every other SC citation in the file re-checked for conformance level** — no repeat of the 2.4.13 error.

### Both blockers fixed

**1. DOET chapter.** The gulfs are ch. 2 (*The Psychology of Everyday Actions*), not ch. 1. Same byte count, headroom unchanged at 6. Worth naming what this was: the house rule is *cite the chapter, not the count* — written because DOET carries two different lists in two chapters — and the first thing that rule produced was a wrong chapter.

**2. The ledger tests.** Two characterization tests hard-asserted the ledger was **empty**, so this batch's two legitimate declarations failed them and took the mutation harness's control down too.

Not repinned to "expect two rows" — that just moves the breakage to the next declaration, and the test's own docstring already anticipated rows being added (*"if a row is ever added, it was added by a change that needed it"*). Repinned to the property that matters: **the ledger parses, and every row names a real skill with a positive count.** A malformed row is the failure worth catching — it reads as a declaration to a human and as nothing to the parser, which puts the removal back to being silent, the whole defect this gate exists for.

### Honest note on local verification

The scratch venv broke during downtime (`pluggy.__file__` is `None` at conftest import) and **`origin/dev` fails identically**, so it is the environment rather than this branch. CI's own environment ran it: **test=SUCCESS**, all six checks green.

### Non-blocking, filed rather than folded in

The panel found `token-naming-conventions/SKILL.md:49` still says every value "must clear an APCA/WCAG check" — the same both-required framing #170 retires elsewhere. An agent loading it for a naming task would block a token that passes APCA and fails WCAG, contradicting `apca-contrast`'s new instruction. Filed separately.

Issue #252, opened 2026-08-19
Design-set audit from a consumer repo: the marketing carve-out, the substrate gap, and seven smaller findings


thrillmot:
A refute-first audit of the 18 design skills as installed in a consumer repo (all subscription pins verified with the real folder-hash algorithm — 0 drifted, control-tested). Findings ranked; quotes verbatim.

## 1. Five skills carve out "marketing" as outside the system — wrong in kind

- `web-interface-guidelines-review:22` — "Marketing-only landing pages with bespoke design that intentionally lives outside the system."
- `spacing-system:19` — "One-off marketing surfaces where the layout is hand-tuned and not part of the system's grid."
- `component-sizing-principles:20` — "One-off marketing CTAs designed for a specific page (e.g. hero buttons at 72 px)."
- `line-height-grid:19` — "One-off display elements (hero marketing splashes) where … the grid doesn't apply."
- `oklch-color-space:19` — "a marketing landing-page splash"

The scope test gates on **surface genre** when it should gate on **whether a system exists**. A marketing page built in a token-driven system is in the system — and a consumer whose entire deliverable is a marketing site gets five skills standing down at once. Proposed fix: replace each carve-out with the test `design-system-consistency` already uses correctly — "no token layer on disk."

## 2. No skill carries the substrate rule

`design-system-consistency` #6 flags a hand-rolled control only "where a system component exists." In a substrate-disciplined codebase the violation stands even when the system lacks the part — the answer is propose-don't-hand-roll. `web-interface-guidelines-review` §6 audits a hand-rolled overlay's Escape behavior rather than flagging the hand-roll. No existing skill is the natural home; this wants an amendment to `design-system-consistency` or a new skill.

## 3. `designing-elite-ui` is the structural outlier

Zero of the house's five sections; a `## Deploying This Skill` section addressed to the author, paid for on every load; `**REQUIRED COMPANION:** orchestrating-elite-agent-qa` as a bare unlinked name; and its worked example is a hand-rolled "build-free vanilla JS" dashboard — modelling exactly what substrate-disciplined consumers forbid. It is REQUIRED BACKGROUND for `visual-polish`, so the shape propagates.

## 4. Two owners of one floor

`component-sizing-principles` defers the 24px floor to `spacing-system` and then restates it, plus both restate the icon ladder `12, 16, 24, 32, 40, 48`. One owner, one statement.

## 5. Unsourced numbers

[...]

Issue #189, opened 2026-08-07
placement-map.json omits skdd from four default-on subscribers arrays, and skdd now commits a lock saying otherwise


thrillmot:
Raised from `skdd`, per SPEC §5.3 — a change requested of another repository belongs as an issue here rather than a commit from there.

## Measured

`docs/placement-map.json` at `main` lists these four with `distribution: default-on` and an identical `subscribers` array:

```
critical-issues-only, evidence-based-review, respect-existing-conventions, clud-bug-collaboration
  subscribers: ["clud-bug", "clud-bug-app", "logmind", "tokenomics", "protocol", "reporulez", "rezgen"]
```

`skdd` is not among them. It holds all four, byte-identical to this catalog, and as of thrillmade/skdd#32 it commits a `skills-lock.json` declaring them as subscriptions of `thrillmade/agent-skills` at ref `2dc8360c8bdfb916d37fc85c23b04f63559daad6`.

Identity was proven twice. The four git blob SHA-1s on `skdd`'s `origin/dev` equal the four here at `2dc8360`:

| skill | blob SHA-1 (both sides) | sha256 of raw bytes (both sides) |
|---|---|---|
| `clud-bug-collaboration` | `06d1807425384399639a6d2a866d0690b4650b35` | `247a443753077b98b01b45c915bea6a50a8b257404c08a7f796f6dd327e4e84c` |
| `critical-issues-only` | `db392b2a7df25ad1a210766fa291e5153b58c1ff` | `0527da0fc07dc7f6654492d4c7cee3e17fd96332e6877c977ce4905cec3d76bc` |
| `evidence-based-review` | `a8a4504365d67d76b7c3c0de0670a092bede5071` | `f7b9b55722000f3b5ff91a64d48ce89f09d3b96e14fa8b5b1a598245ec049089` |
| `respect-existing-conventions` | `e18d3516d7480382f98ceda3f9bef56a4aee0280` | `936899cf6899179023b18ee844d3c9996210eae6bd8b801f750f70e97578123b` |

Reproduce either side:

```
gh api "repos/thrillmade/agent-skills/contents/skills/evidence-based-review/SKILL.md?ref=2dc8360c8bdfb916d37fc85c23b04f63559daad6" --jq .sha
[...]

Issue #183, opened 2026-08-01
skill-census.yml files census issues as github-actions[bot], not the steward App — flip GH_TOKEN once Issues R/W is granted


thrillmot:
Filed from the skdd lane per SPEC §5.3.

## What the workflow already says to do

`.github/workflows/skill-census.yml` mints the steward App token (App renamed `thrillmade-orchestrator` → `skdd-steward`, same App ID/key/installation per `logmind`'s `AGENTS.md`), scoped to 8 repos, at lines 90-106 (`origin/main`):

```yaml
      - name: Mint steward App token
        id: app_token
        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3
        with:
          app-id: ${{ secrets.THRILLMADE_ORCHESTRATOR_APP_ID }}
          private-key: ${{ secrets.THRILLMADE_ORCHESTRATOR_PRIVATE_KEY }}
          owner: thrillmade
          repositories: |
            clud-bug
            clud-bug-app
            logmind
            tokenomics
            protocol
            reporulez
            rezgen
            skdd
```

But the two steps that actually post — the panel step (reads the `census` docket via `gh issue list`) and the filing step (posts the digest + verdict issues) — both use the workflow's own default token instead:

```yaml
# line 144, "Census panel (judgment)":
          GH_TOKEN: ${{ github.token }}

# line 161, "File issues (digest + verdicts)":
          GH_TOKEN: ${{ github.token }}
```

The workflow's own comment immediately above the filing step's token, lines 158-160, says why that's temporary:

```yaml
          # Filing uses GITHUB_TOKEN (github-actions[bot]); switch GH_TOKEN to
          # the App token once the App is granted Issues R/W to file under the
          # steward identity.
```

(Line numbers verified directly against `origin/main:.github/workflows/skill-census.yml`, current as of this issue.)

## Why it matters to skdd

§5.4 requires the steward to act under its own identity:

> The steward acts under its own identity, with a short-lived token minted per run rather than a personal one, and one installation per organisation.

[...]

PR #255, opened 2026-08-23
The marketing carve-out, the substrate rule, and a routing map blind to its two biggest consumers (#252, #189)

thrillmot:
Tranche A1+A2 and B1 of the launch plan. Batched because they are disjoint file sets and neither depends on the other.

## A1 — five skills stood down on "marketing", which is wrong in kind

From the consumer audit in **#252**. Each gated on surface **genre** when the test should be whether a **system exists**. A marketing page built in a token-driven system *is* in the system — and a consumer whose whole deliverable is a marketing site got five skills standing down at once. **This studio builds marketing sites.**

`design-system-consistency` already had the correct test at `:36` — **"No system on disk"**. Each carve-out now opens with the same phrase and finishes in that skill's own vocabulary:

| skill | now reads |
|---|---|
| `web-interface-guidelines-review` | "No token layer on disk — this checklist has nothing to check against. A marketing page built inside one is still in scope." |
| `spacing-system` | "…no unit primitives declared anywhere, so there is no grid to be off of." |
| `component-sizing-principles` | "…no curated ladder to be off of." |
| `line-height-grid` | "…no grid declared, so there is nothing to snap to." |
| `oklch-color-space` | "…no catalog for this color to enter." |

## A2 — no skill carried the substrate rule

In a substrate-disciplined codebase a hand-rolled control is a violation **even when the system lacks the part**. `design-system-consistency` #6 gated on *"where a system component exists"*, so it missed exactly the case that matters. Amended rather than adding a skill — it already owns the adjacent rule:

> A hand-rolled button, card or input — **the finding stands whether or not a system component for it exists.** Where one exists it's visible as subtly different padding, radius, hover or focus ring from its siblings; fix: the component, not corrected values. **Where none exists, propose it upstream — a missing part is never license to hand-roll one.**

## B1 — the routing map was blind to its two largest consumers

`placement-map.json` declared 8 subscribers. Measured on disk:

```
arlyn-working    23 skills pinned    appeared in the map 0 times
arlyn-delivery   17 skills pinned    appeared 0 times
                                     (control: 'logmind' -> 9)
```

Every mechanism reading that file — the census, any future fan-out — was blind to the repos the work exists for. **23 subscriber arrays corrected**, each against its justifying lock: 17 entries got both repos, 6 got `arlyn-working` only. Locks verified by ancestry — arlyn-working's is on `main`, arlyn-delivery's is on `dev` and not yet promoted, and the report says which.

A suspected `tokenomics` discrepancy was **re-checked and refuted** before any edit: the apparent vercel-only content belonged to a gitignored local artifact in a different repo, not the committed lock.

## Two judgment calls, surfaced rather than buried

**1. Dispatchers as standing subscriptions.** 17 of the 23 corrections are dispatchers or lenses whose `notes` say *"a repo consumes what it routes to, not the dispatcher itself as a standing subscription."* Both arlyn repos pin them directly anyway. Reconciled to the lock per the brief, and the notes left alone — but **either that note is wrong or those repos subscribed to something they shouldn't have**, and this PR does not settle which.

**2. `CONSUMER_REPOS` deliberately not touched.** Adding the arlyn repos there is not a free one-liner: `skill-census.yml` mints an App token against a hardcoded `repositories:` allowlist that is the same 8 repos. A repo added to `CONSUMER_REPOS` without a matching grant produces a **permanent weekly `error` row instead of real data** — the same trap the existing `finding-a-catalog-skill` note already names. Left as an implication for the #183 grant.

## The byte problem this creates, and it is now acute

| file | headroom after |
|---|---|
| `design-system-consistency` | **4 bytes** |
| `web-interface-guidelines-review` | **19 bytes** |

Both are now effectively unmaintainable — the next editor must cut before adding. No prose was cut to fit here (`check_prose_retention` clean, ledger untouched, every replacement equivalent-or-larger), but this is the sharpest evidence yet for **clud-bug#305**: whether their reviewer reads a `references/` subdirectory decides if these skills can ever grow again.

## Gates

```
validate_skills.py         link scope: 370 relative / 51 absolute
                           OK: 56 skills validated cleanly              EXIT=0
check_prose_retention.py   OK: 6 changed SKILL.md file(s).              EXIT=0
```

`pytest` skipped locally — the scratch venv is broken (`pluggy.__file__` is `None`) and **`origin/dev` fails identically**, so it is environmental. CI runs its own.

### DIFF
--- a/docs/decisions-branches/fix__marketing-carveout-and-substrate.md
+++ b/docs/decisions-branches/fix__marketing-carveout-and-substrate.md
@@ -1,0 +1,14 @@
+← back to [docs/timeline.md](../timeline.md)
+
+## 2026-08-23 16:14 - Replace the marketing-genre carve-out with the system-exists test, and make design-system-consistency's re-implementation rule cover the substrate gap
+
+**Reasoning:** Issue #252: five skills (web-interface-guidelines-review, spacing-system, component-sizing-principles, line-height-grid, oklch-color-space) stood down on surface genre ('marketing') when the real test is whether a token layer exists on disk -- a marketing page built inside a token-driven system is in the system, and a studio whose whole deliverable is marketing pages was getting five skills standing down at once. design-system-consistency already had the correct test ('No system on disk'); ported it to each skill's own voice. Separately, design-system-consistency rule 6 only flagged a hand-rolled control 'where a system component exists', missing the substrate-disciplined case where the violation stands even when the system lacks the part -- amended rule 6 to cover both branches (propose upstream, don't hand-roll, regardless of whether the part exists yet).
+
+**Alternatives considered:** Add a new skill for the substrate rule instead of amending design-system-consistency -- rejected per the issue's own instruction: it already owns the adjacent rule., Move the carve-out fixes to references/ subdirectories for more room -- rejected: references/ is blocked pending clud-bug#305 and unread by the CI reviewer anyway.
+
+**Implications:**
+- web-interface-guidelines-review now has 18 bytes of headroom against the 8192-byte body cap, and design-system-consistency has 3 -- any future edit to either file needs to cut before it adds.
+- The shared phrase 'No token layer on disk' now opens the carve-out bullet in all six design skills that gate on system existence, which is worth keeping consistent if any of them changes wording again.
+
+---
+
--- a/docs/decisions-branches/fix__placement-map-reality.md
+++ b/docs/decisions-branches/fix__placement-map-reality.md
@@ -1,0 +1,13 @@
+← back to [docs/timeline.md](../timeline.md)
+
+## 2026-08-23 16:16 - Reconcile placement-map.json subscribers against real skills-lock.json files: add arlyn-working and arlyn-delivery, the two largest real consumers, to every skill entry their locks actually pin
+
+**Reasoning:** docs/placement-map.json declared 8 subscribers but the census cannot see arlyn-working (23 pinned skills, committed on main) or arlyn-delivery (17 pinned skills, committed+pushed on dev) at all; every downstream fan-out (weekly census, future placement checks) was silently skipping the two largest real consumers
+
+**Alternatives considered:** leave the map as-is and treat the 8-repo census scope as the whole picture -- rejected, the map's job is to describe reality that other tooling reads, and it was already blind to 40 real subscriptions across 23 skill entries
+
+**Implications:**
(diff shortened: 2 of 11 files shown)


thrillmot:
## Panel: MERGE WITH FIX → both blockers fixed → merging

The panel caught a **real regression I would have shipped**, and it is the useful kind: locally true, contextually wrong.

### 1. WIG must not stand down for a missing token layer

WIG's checklist covers **verb-noun labels** and **link-vs-button** — zero token dependency — and WIG is their **sole owner** (`reviewing-design-work` only references it). The panel's failure scenario: a repo with no token layer ships an anchor-as-button with a bad label, WIG stands down entirely, both violations go uncited.

So **neither test belonged there** — not the old marketing genre, not the new token layer. The bullet now says what is true:

> No carve-out for a missing token layer: only section 6 skips without one.

It also **frees 51 bytes**, taking WIG from 19 bytes of headroom to 70 — so the file that was second-tightest in the catalog is now breathable.

The panel's discriminating detail is worth keeping: the sibling four are safe precisely *because* they scope their justification to their own domain ("no unit primitives", "no curated ladder", "no grid", "no catalog"), so a repo with colour tokens but no spacing scale still stands `spacing-system` down and keeps `oklch-color-space` in scope. It constructed that case and confirmed it resolves correctly.

### 2. A wrong number baked into a decision record

The decision file claimed **17** dispatcher/lens entries carrying the "consumes what it routes to" note. Measured: **3**. The others are clud-bug#243 authoring-home notes — about *where edits happen*, not about subscription — and folding them in generously still only reaches 8. Corrected in the file, with the measurement.

A wrong number in a decision record is worse than in a PR body, because this repo treats those as a source of truth.

### Not taken, with the reason

The advisory to add "filed once" dedup language to rule 6 is real, and it is the same gap rule 2 already patched for itself. But it is a behaviour change to a rule this PR only just amended, and it belongs in its own change where it can be reviewed on its own terms. Filing it.

### On the judgment call the PR deliberately left open

The panel took a position and I agree: **the note is the stale artifact, not the repos.** Skills are consumed via static `skills-lock.json` pinning, not dynamic browsing — an agent in `arlyn-working` needs `consuming-a-design-system`'s own file on disk to ever be routed to the L0 primitives it names. "Consumes what it routes to" describes an idealised footprint that does not match how distribution actually works here.

### And a correction to something I have been repeating

I have been reporting the broken local venv as environmental and universal. **The panel ran the full suite in a fresh venv and it passed** — 907 fast subset, 128 across both mutation files. So it is specific to the tree's scratch venv, not to the machine. I should have tested that rather than generalising from two identical failures.

PR #257, opened 2026-08-24
Finish the design-set audit (A3–A7), and fix #183 — whose premise was false

### LINKED ISSUE #183: skill-census.yml files census issues as github-actions[bot], not the steward App — flip GH_TOKEN once Issues R/W is granted

thrillmot:
Closes the consumer audit in **#252** and fixes **#183**. Batched — disjoint files, neither depends on the other.

## #183 — the premise was false, and I had been repeating it

I reported this to the CEO for days as *"needs your Issues R/W grant"* and called it the gate on the whole notify/sync loop. Measured:

```
gh api /apps/skdd-steward --jq '.permissions.issues'   ->  write   (granted 2026-07-17)
agent-skills in the census App-token repo list         ->  0       (control: logmind -> 1)
```

**The App has had the permission since July.** The blocker was **token scope, not permission** — `skill-census.yml` mints against a hardcoded repo list omitting this repository, so filing fell back to `github.token` and posted as `github-actions[bot]`. The file's own comment claiming *"the App has no Issues permission at all"* was wrong.

It had already been debunked **in that issue thread on 2026-08-01**. I recycled it without reading the thread, put it in the roadmap, and told the CEO. `skdd#9` cites this file by line, so a cross-repo mechanism was described as blocked on a grant that already existed.

Three edits: add the repo, flip both `GH_TOKEN` lines, correct the comment. Install scope confirmed separately — `repository_selection=all`, `issues=write` at the org level.

**The risky question, answered with evidence rather than assumed:** a minted App token is scoped by both the App's permissions and what the step requests, so a missing `permission-issues` input would produce a runtime 403 — worse than the status quo. Fetched the action's own `action.yml`: *"By default, the token inherits all of the installation's permissions."* No input needed.

Then reduced privilege unprompted: with nothing left using `github.token` for issues, the job's `permissions:` drops to `contents: read`.

**Fails loudly, verified empirically not reasoned.** New `tests/test_skill_census_token_failure.py` extracts and executes the real embedded bash against a fake `gh` simulating a 403. Mutation-tested by reintroducing a silent `|| echo fallback` around the digest create — test went red, restored, green.

**Unverified until it runs:** that the token actually *writes*. A `workflow_dispatch` dry-run confirms the mint and the read but never reaches a mutating call. Real proof is the next scheduled run (Mon 12:00 UTC) or a manual non-dry run that files a real issue.

## A3 — `designing-elite-ui` was the structural outlier

Zero of the five house sections; an author-addressed `## Deploying This Skill` paid for on every load; a bare unlinked `**REQUIRED COMPANION:**`; and **a worked example modelling a hand-rolled "build-free vanilla JS" dashboard** — exactly what the substrate rule landed in #255 forbids. It is REQUIRED BACKGROUND for `visual-polish`, so the shape propagated.

All five sections added, author section cut, companion is a real link, and the example now composes from an existing token-driven system. Tail order chosen as `Verification | Sources | Cross-references` **because 3 of its 4 `design-critic-lenses` siblings already use it** — matching, not inventing. Headroom 2867 → 460.

## A4 — one owner per fact

`spacing-system` sole-owns the WCAG 2.5.8 24px floor; `component-sizing-principles` sole-owns the icon ladder `12, 16, 24, 32, 40, 48`. Each links to the other instead of restating. **Net freed 173 bytes.**

## A5 — cite or cut

`line-height-grid`'s "tighter than 1.20 clips diacritics" and "1.45 fatigues the eye" — **verified against WCAG 2.2 SC 1.4.12, which requires ≥1.5× and says nothing about 1.2 or 1.45.** Cut both, added the citation for the ratio it does support.

`spacing-system`'s radius formula reframed to two universal endpoints plus *"rung count and growth rate is the system's own personality call"*, deferring the ladder to `udts-spacing-defaults`. The 72px example was already gone — removed by #255, confirmed by grep, no edit needed.

## A6 — measured, deliberately not changed

The tail-order ruling is the CEO's. **No file's order was changed to converge on anything.** Fresh reproducible counts over this catalog:

| scope | counts |
|---|---|
| design dispatchers + primitives (18) | `(V,C,S)`×8 · `(C,V,S)`×6 · Cross-refs only×3 · `(V,S)`×1 · **0 no-tail** |
| + critic lenses (23) | `(V,C,S)`×8 · `(C,V,S)`×7 · `(V,S,C)`×4 · Cross-refs only×3 · `(V,S)`×1 |
| full catalog (56) | `(V,C,S)`×14 · `(C,V,S)`×9 · Cross-refs only×6 · `(V,S,C)`×5 · 19 no-tail |

**These do not match the issue's "7/5/4 installed; 9/9/4 upstream"** — the lane could not reach the external consumer repo the audit measured, so these are counts over *this* catalog and are stated as such rather than reconciled by assumption.

## A7 — `consuming-a-design-system` §5

"Alias, don't fork" reframed as a **fenced interim**: issue-linked, deleted when the upstream token lands. A consumer running fenced-interims discipline previously read it as license for permanent drift.

## Prose removals — declared, not cut-to-fit

Two, both in files with large headroom: `component-sizing-principles` (19 words, the duplicated WCAG floor) and `spacing-system` (55 words, the unsourced radius formula). `design-system-consistency` sits at 3 bytes and **was not touched**.

## Gates

```
validate_skills.py         OK: 56 skills validated cleanly       EXIT=0
check_prose_retention.py   OK: 5 changed SKILL.md file(s).       EXIT=0
pytest (fresh venv)        907 fast subset / 1035 full           EXIT=0
```


thrillmot:
## Panel: MERGE WITH FIX → fixed → merging

All four census claims **confirmed independently at source level**, not taken on report. The panel read `dist/main.cjs` at the pinned SHA (verified as the real v3.2.0 tag, not a stale pin) and found `getPermissionsFromInputs` returns `undefined` with no `permission-*` set, so the `permissions` key is omitted from the token request entirely — cross-checked against GitHub's OpenAPI text and the **live** installation (`repository_selection: all`, `issues: write`). Three independent confirmations that the flip works at runtime.

It also **mutation-tested the new test itself**: wrapped the digest `gh issue create` in `|| true` and watched `test_bad_token_fails_loud_not_silent` go red. The test has teeth.

### The finding, and it is about the gate rather than this PR

's `## Deploying This Skill` section — 55 words — was genuinely deleted, and **`check_prose_retention.py` returned exit 0.**

Not a bug in the change; a **scoping blind spot**. The gate scores net words across three coarse buckets per file (frontmatter / prose / code), not per heading, so the five house sections added in the same pass offset the deletion inside the same prose bucket.

That is worth more than the instance: the gate exists because *"three skills lost real content to a link-conversion sweep… and every check went green."* **A restructure is exactly that shape** — adds and removes in one pass, same bucket, nets out. So it is strongest against a pure deletion and weakest against the case that most resembles it in effect. Filed as **#258** with three candidate fixes.

**Fixed here:** the ledger row, added. The cut itself stands — that section addressed the skill's *author*, not its reader, and was paid for on every load. What was missing was the row. This PR declared two other cuts in the same change, and **a disclosure policy applied only when a gate forces it is not a policy.**

### Also confirmed by the panel, independently

- **A5's cuts were correct** — it fetched the live W3C text: SC 1.4.12 requires ≥1.5× and mentions neither 1.2 nor 1.45. No external source supports the removed claims either.
- **A4 orphaned nothing** — each fact has exactly one owner and the other genuinely links; the icon ladder was already sole-owned pre-PR, so only the duplicate went.
- **A6 changed no ordering** — heading order is identical in all four other files; only `designing-elite-ui` gained sections, and its pre-existing four kept their relative order.
- **Full suite 1037 passed** from a freshly built venv, with the zero-failures claim itself control-tested.

Unverified and stated as such: whether the App token actually **writes** in production. Static analysis and live API queries go as far as they can; real proof is Monday's scheduled run.

thrillmot:
## Panel: MERGE WITH FIX → fixed → merged

All four census claims **confirmed independently at source level**, not taken on report. The panel read `dist/main.cjs` at the pinned SHA (verified as the real v3.2.0 tag, not a stale pin) and found `getPermissionsFromInputs` returns `undefined` with no `permission-*` set, so the `permissions` key is omitted from the token request entirely — cross-checked against GitHub's OpenAPI text and the **live** installation (`repository_selection: all`, `issues: write`). Three independent confirmations that the flip works at runtime.

It also **mutation-tested the new test itself**: wrapped the digest `gh issue create` in `|| true` and watched `test_bad_token_fails_loud_not_silent` go red. The test has teeth.

### The finding, and it is about the gate rather than this PR

`designing-elite-ui`'s `## Deploying This Skill` section — 55 words — was genuinely deleted, and **`check_prose_retention.py` returned exit 0.**

Not a bug in the change; a **scoping blind spot**. The gate scores net words across three coarse buckets per file (frontmatter / prose / code), not per heading, so the five house sections added in the same pass offset the deletion inside the same prose bucket.

That is worth more than the instance: the gate exists because *"three skills lost real content to a link-conversion sweep… and every check went green."* **A restructure is exactly that shape** — adds and removes in one pass, same bucket, nets out. So it is strongest against a pure deletion and weakest against the case that most resembles it in effect. Filed as **#258** with three candidate fixes.

**Fixed here:** the ledger row, added. The cut itself stands — that section addressed the skill's *author*, not its reader, and was paid for on every load. What was missing was the row. This PR declared two other cuts in the same change, and **a disclosure policy applied only when a gate forces it is not a policy.**

### Also confirmed by the panel, independently

- **A5's cuts were correct** — it fetched the live W3C text: SC 1.4.12 requires ≥1.5× and mentions neither 1.2 nor 1.45. No external source supports the removed claims either.
- **A4 orphaned nothing** — each fact has exactly one owner and the other genuinely links; the icon ladder was already sole-owned pre-PR, so only the duplicate went.
- **A6 changed no ordering** — heading order is identical in all four other files; only `designing-elite-ui` gained sections, and its pre-existing four kept their relative order.
- **Full suite 1037 passed** from a freshly built venv, with the zero-failures claim itself control-tested.

Unverified and stated as such: whether the App token actually **writes** in production. Static analysis and live API queries go as far as they can; real proof is Monday's scheduled run.

Issue #256, opened 2026-08-23
design-system-consistency rule 6 has no 'filed once' dedup, unlike rule 2 directly above it


thrillmot:
Advisory finding from the panel on #255, deliberately not folded into that PR.

Rule 2 carries explicit anti-noise language — findings are *"filed once — not one finding per hue you would have picked differently."*

**Rule 6 does not**, and #255 just widened it: the finding now stands whether or not a system component exists, with *"where none exists, propose it upstream."* In a repo with a thin design system and many hand-rolled buttons, a literal reading produces **one finding per instance** instead of one finding for the missing component class.

Bounded, not unbounded: rule 6 only fires under the skill's own precondition (`## When NOT to use` → *"No system on disk. File one finding … and stop."*), so it cannot misfire in a genuinely token-less repo. This is about volume inside a repo that *does* have a system, not about false positives.

The fix is the precedent rule 2 already set — one finding per missing component class, not per instance.

**Why it was not fixed in #255:** it is a behaviour change to a rule that PR had only just amended, and bundling it would have meant the widening and the bounding were reviewed as one thing. Byte cost is the constraint — `design-system-consistency` is at **4 bytes** of headroom, the tightest file in the catalog, so this needs a trim to land at all. That makes it a real argument for clud-bug#305.

thrillmot:
Fixed in #263, merged to `dev`. Verified on the merged branch:

```
$ gh api "repos/thrillmade/agent-skills/contents/skills/design-system-consistency/SKILL.md?ref=dev" \
    --jq .content | base64 -d | grep -c 'propose it upstream, filed once'
1
```

Rule 6 now carries rule 2's precedent: filed once per missing component class, not one finding per hand-rolled instance.

**The interesting part is where the bytes came from.** This file had **3 bytes** of headroom (re-measured; the tracked figure of 4 was off by one — and an independent panel then found `test-discipline` at **1 byte**, so "tightest in the catalog" was mine to retract). Room was bought by cutting a `Sources` bullet that restated three things already stated in full elsewhere — all three located and quoted by the panel, so it is dedup rather than content loss — and **declared in `docs/prose-removals.md`**, not removed silently.

Headroom 3 → 139 bytes.

Issue #253, opened 2026-08-19
token-naming-conventions still requires both APCA and WCAG, contradicting #170's retired both-models gate


thrillmot:
Found by the panel on #251, correctly not blocked on.

`skills/token-naming-conventions/SKILL.md:49` reads that token values "carry a contrast-pairing obligation — every value must clear an APCA/WCAG check".

#170 retired exactly that framing in `apca-contrast` and `wcag-contrast`, which now treat WCAG as **optional, off-by-default** — because forcing a P3-native palette to satisfy an sRGB-era threshold blocks tokens that are correct.

**Failure scenario:** an agent loads `token-naming-conventions` for a naming task, reads "every value must clear an APCA/WCAG check", and blocks a token that passes APCA and fails WCAG — directly contradicting the instruction `apca-contrast` now gives it. Two skills in one catalog telling an agent opposite things.

`design-token-naming` carries the same old phrasing but is explicitly SUPERSEDED and frozen, so it is out of scope.

Not caught by #170's own lane because its decision log scoped the sync to `apca-contrast` and `wcag-contrast` keeping pace with *each other*, and nothing checks a policy change against skills that merely *cite* the policy. That gap is the more interesting half of this issue.

thrillmot:
Fixed in #263, merged to `dev`. Verified on the merged branch:

```
$ gh api "repos/thrillmade/agent-skills/contents/skills/token-naming-conventions/SKILL.md?ref=dev" \
    --jq .content | base64 -d | grep -c 'WCAG 2.2 AA is an optional'
1
```

APCA Lc is the gate; WCAG 2.2 AA is an explicit, off-by-default cross-check — matching `apca-contrast` instead of contradicting it.

**The more useful half was checking whether anything else still cited the retired policy.** A catalog-wide search found exactly one other hit: `design-token-naming`, which is `SUPERSEDED` and frozen during the tokenomics window, and is out of scope by this issue's own terms. Every other skill naming both (`designing-a-design-system`, `oklch-color-space`, `reviewing-design-work`, `web-interface-guidelines-review`) already had the correct framing. So #170's sync missed nothing beyond this one file.

Headroom 657 → 560 bytes.

Issue #250, opened 2026-08-18
skill-census.yml swallows a failed 'gh issue list' with no annotation — the recurring silent-skip shape, twice in one file


thrillmot:
Found by the panel on #249, which correctly declined to block on it.

## The shape

```bash
PRIOR_DIGESTS=$(gh issue list ... 2>/dev/null | jq ... ) || PRIOR_DIGESTS=""
```

If `gh issue list` fails — rate limit, token expiry, transient API error — the fallback swallows it. **No `::warning::`, no `::error::`, close block silently skipped, step exits 0.** A malformed-but-successful response is slightly better: `jq: parse error` reaches the raw step log, but still produces no GitHub annotation and still exits 0.

Reproduced by the panel with a scripted mock `gh` returning exit 1 and empty stdout.

## Why it was not a blocker on #249

Three things, all established rather than assumed:

1. **It fails safe, not dangerous.** It can only cause a close to be *skipped*, never a wrong issue to be closed.
2. **It self-heals.** The guard has no same-cycle requirement, so any later successful run closes the same stale digest.
3. **It is pre-existing, not introduced.** The identical idiom is already on `dev` in the `EXISTING_DIGEST` lookup (`skill-census.yml`, the digest-lookup step). #249 followed the file's convention.

## Why it is still worth fixing

**It is this repo's most-repeated defect**, and the file now contains it twice. The pattern:

> a mechanism that cannot report its own uselessness will be ignored before it is fixed

A census whose issue-listing quietly fails looks *identical* to a census with nothing to close. That is the same failure as `check-links` reporting green over files it never read (#234), the size gate passing over prose it never compared (#245), and the versions index that would have reported `current` for a skill it had never seen (#238) — all closed this week, all the same shape.

## The fix

Add `::warning::` on both fallbacks, matching the annotations the same step already emits for a missing `census_key` and an unknown label. That is the convention; these two sites are the exception.

Small, and it makes the difference between *"nothing to do"* and *"I could not find out"* legible — which is the distinction every one of the defects above turned on.

thrillmot:
Fixed in #263, merged to `dev`.

A failed `gh issue list` was discarded, so a lookup outage was indistinguishable from "nothing filed" — and the census would refile. Verified on the merged branch:

```
$ gh api "repos/thrillmade/agent-skills/contents/.github/workflows/skill-census.yml?ref=dev" \
    --jq .content | base64 -d | grep -c '::warning::'
6
```

[...]

PR #263, opened 2026-08-24
Five issue fixes in one pass: the prose gate's blind spot, a swallowed census failure, a drift guard that could not drift

### LINKED ISSUE #258: The prose gate scores three coarse buckets per file, so added sections silently pay for deleted ones
Found by the panel on #257, which walked through it in good faith and then said so.

## The blind spot

`check_prose_retention.py` scores **net word count across three coarse scopes per file** — frontmatter, prose, code — **not per heading**. So words added anywhere in the prose bucket offset words deleted anywhere else in it.

**Reproduced on real work.** #257 restructured `designing-elite-ui`: it added the five house sections and deleted `## Deploying This Skill` (55 words). The gate returned:

```
check_prose_retention.py --base origin/dev --head <branch>   OK, exit 0
```

A genuine 55-word section vanished and the gate scored it as zero, because the new template sections' words landed in the same bucket. Confirmed the deletion is real with a control: `grep` for `retrieval + application` returns nothing on the branch, while `grep` for `Verification` returns matches — so the absence is real, not a grep failure. That phrase existed in **only that file** catalog-wide.

## Why this matters more than the one instance

The gate exists because *"three skills lost real content to a link-conversion sweep that needed to fit under the 8192-byte size gate, and every check went green."* **A restructure is exactly that shape**: it adds and removes in one pass, in the same bucket, and nets out.

So the gate is strongest against the case it was built for — a pure deletion — and weakest against the case that most resembles it in effect.

## What it is not

Not a false green in the dangerous direction *by design intent*: the gate's own docs are explicit that reformatting is free and that scoping is per-part. It is doing what it says. The gap is that "part" is coarser than an author's mental model of a section, and a restructure crosses that boundary invisibly.

## Possible answers, none obviously right

1. **Score per `##` section** rather than per bucket. Most faithful to intent; also the most likely to produce false positives on legitimate reorganisation, which is explicitly free today.
2. **Flag large offsetting churn** — a file where both added and deleted word counts are large but the net is small — as *needs a declaration*, without asserting which passage went.
3. **Leave it and rely on disclosure discipline.** #257 disclosed the cut in its commit message and PR body, just not in the ledger, and the row was added on review. That is the system working — via a human, not a gate.

Option 2 is probably the honest middle: it says *"something substantial moved here, tell me why"* without pretending to know what.

## Related

This is the fourth member of the family the catalog keeps finding — a mechanism that reports success over the case it cannot see. See #234 (a required gate that read no skill bodies), #245 (the same gate never comparing *added* files), #250 (a silent skip on a failed lookup). Three are closed. This one is subtler than all of them, because the gate genuinely runs and genuinely measures — it just measures at the wrong grain.

thrillmot:
Closes #258, #250, #253, #256. Advances #221 and #217 without closing either.

Batched because the findings overlap in files and reviewer attention — not because they share a cause. Five separate PRs would mean five reconciliations of the same files plus review overhead without review value.

## #258 — the prose gate could be paid off

`check_prose_retention.py` scored three coarse buckets **net**, so a deletion could be covered by a bigger addition in the same scope.

The founding case is real and now vendored as a fixture: `designing-elite-ui`'s actual #257 restructure (`602bb6c`→`8896dcc`) **cut 36 words and scored a net gain — exit 0.** It now fires:

> 36 words that nothing in the same part of it replaced… Counted by net alone its prose would not have fired: 328 words arrived there in the same change.

### The threshold is chosen, not discovered — and the code says so

This is the part worth reviewing hardest. Calibrating `CHURN_FLOOR` by replaying the signal over the same corpus `FLOOR`'s comment measures found a **dense, continuous distribution** (1, 2, 4, 5, 6 … 95, 100 — no gap), unlike `FLOOR`, which has a genuine empty band. Worse, the founding case (36 words) sits *below* several presumably-legitimate large rewrites in the same history (the #231 lens-fill, 95–100).

A magnitude-only cutoff cannot cleanly separate "restructure hiding a cut" from "wholesale legitimate rewrite." A retained-fraction discriminator was also tried (0.84 vs 0.33 — a real split) and rejected as unlikely to generalise.

So `CHURN_FLOOR=20` is **documented in the code as a chosen order-of-magnitude margin above `FLOOR`, not a discovered gap.** It fires on 19 distinct (path, scope) pairs across the whole catalog history — not on every restructure. Recording that honestly matters more than the number.

### It immediately found a second instance — in the commit that fixed the first

Applied to `dev`'s own history, the new detector flags `spacing-system`'s icon-ladder move to `component-sizing-principles` in commit `d3963dd4` — masked by a border-radius paragraph that replaced a code block in the same change, and **not covered by that commit's ledger row**, which declares only the code-scope cut.

Outside this branch's diff, so CI here is unaffected. Filed rather than silently fixed: it is a content judgment for the skill owner, not a gate-engineering one.

**Mutations** (hand-applied, restore-verified byte-identical):

| mutation | result |
|---|---|
| disable `self.churn` | 2 red (`{} == {'prose': 36}` / `{'prose': 21}`) → restored → 385 green |
| `self.net` ignores churn contributions | 2 red (`0 == 36`) → restored → 385 green |

Full 89-mutation suite: **89/89 green** (456s).

## #250 — a swallowed census failure read as "no prior issue"

`skill-census.yml` discarded a failed `gh issue list` at the two sites #250 names, so a lookup outage was indistinguishable from "nothing filed" — and the census would refile. One `::warning::` each, matching the file's existing convention.

The two sites #250 **didn't** name (`KEY_MATCH`, `EXISTS`) have the same shape and were deliberately left alone rather than scope-creeping. Flagged below.

New `tests/test_skill_census_gh_list_failure.py` executes the **real embedded workflow step** against a fake `gh` that fails each lookup in turn. Mutation: strip both warnings → both tests red → restored → 10 green.

## #221 — a drift guard that could not detect drift

Not a patch to `docs/file-structure.md` — **that is logmind's generated file, not ours.** Our side is made safe; the rest is reported upstream.

The interesting finding is that the *existing* guard was weaker than it looked: `test_gitignore_covers_local_state.py` compared `.gitignore` against a **hardcoded snapshot list**, so it would stay green forever if `.git/info/exclude` grew an unmirrored 11th pattern — precisely the drift #221 is about.

Both upstream defects were reproduced **live on the installed logmind 1.2.0**, not read off: a mirrored pattern does not leak into `logmind file-structure` output, and — as a control — a pattern excluded only via `.git/info/exclude` **does**. Both cleaned up, `git status` clean after each.

New `test_gitignore_stays_in_sync_with_the_live_info_exclude` reads the live file via `git rev-parse --git-common-dir` (correct from inside a worktree). Mutation: append an unmirrored pattern → red → restored → 3 green.

Upstream report posted to #221. **Issue stays open** — our side is closed, theirs isn't.

## #253 / #256 — two content contradictions

**#253**: `token-naming-conventions` still demanded *both* APCA and WCAG after that framing was retired. Corrected to match: APCA Lc is the gate, WCAG 2.2 AA an explicit off-by-default cross-check. A catalog-wide search for the retired phrasing found exactly one other hit — `design-token-naming`, which is `SUPERSEDED` and frozen during the tokenomics window. So #170's sync missed nothing else. Headroom 657 → 560.

**#256**: `design-system-consistency` rule 6 lacked rule 2's dedup. Now: *"propose it upstream, filed once per missing component class — not one finding per hand-rolled instance."*

That file had **3 bytes** of headroom (re-measured; the tracked figure of 4 was off by one — still the tightest in the catalog). Room was bought by cutting genuine duplication — a `Sources` bullet restating three things already stated in full elsewhere, whose own heading already disclaimed belonging there — and **declared in `docs/prose-removals.md`**, not removed silently. Headroom 3 → 139.

## #217 — the answer was documentation, not code

Nothing needed changing here. `arlyn-working` was read **read-only** throughout (rsync'd to scratch; the live repo's `skills-lock.json` mtime is unchanged). All three things the issue's thread called missing have shipped: the ruling, the subscriber registry (all 23 skills its lock pins), and `skills_current.py`.

Run live against the read-only copy: **30 discovered, 2 current, 21 stale, 5 diverged, 2 local.** The 5 diverged are correctly discriminated, not a tool bug — those slugs didn't exist in the catalog when the ruling was made and were promoted from `arlyn-working` afterward, so "diverged" is the right verdict rather than "stale" or "local."

**Issue stays open** for one real gap: the weekly census still doesn't scan `arlyn-working` / `arlyn-delivery`, so only a subscriber-initiated run can see their drift.

## Merge ordering — read this

This branch collides with **#262** (per-skill identity fields) on three paths:

```
docs/skill-versions.json
skills/design-system-consistency/SKILL.md
skills/token-naming-conventions/SKILL.md
```

**#262 should merge first.** It restamps all 56 skills; this branch then rebases and re-derives stamps mechanically with `stamp_versions.py --write`. The reverse order means hand-reconciling a 56-file restamp against two content edits.

## Verification

Fresh venv (Python 3.11.8 via pyenv — the ambient 3.13 has the known-broken `pip` behind the `pluggy.__file__ is None` symptom). `pytest tests/ -q` → **1046 passed, exit 0**, one run, no exclusions.

## Left for someone else

- #250's `KEY_MATCH` / `EXISTS` sites — same unwarned shape, not named by the issue.
- `spacing-system`'s newly-surfaced masked loss — pre-existing on `dev`.
- The census's App-token allowlist excluding `arlyn-working` / `arlyn-delivery` — deferred as a coupled change on 23 Aug.

## Review posture

Per protocol §8.3 the panel reviewing this is a gate **I** administer — tamper-evident, not independent.


thrillmot:
## Independent adversarial panel — MERGE-WITH-FIXES, **0 blocking**

It rebuilt the calibration replay from scratch rather than trusting the numbers, made a **throwaway shallow clone** to test the CI path, and caught itself reading `$?` after a pipe mid-review. What it attacked and could not break:

- **Fixture provenance** — `git show 602bb6c2ea:…` and `8896dcc1e0:…` diffed **byte-for-byte identical** to the vendored `before.md`/`after.md`. Not hand-tuned to make the test pass.
- **The two edited mutation entries** — ran the *real* subprocess harness against `gains_ignored` and `breakdown_omits_parts_below_their_floor`: `3 passed, 86 deselected`. Both still turn the real suite red on the class of bug they name. Coverage did not die in the refactor.
- **The `self.churn` mutation**, re-run end-to-end and grepped in the tree → 2 red exactly as claimed → byte-identical restore → 5 green.
- **#221's guard under a real CI shape** — cloned `--depth 1` to mimic a runner. `.git/info/exclude` always exists (git's own init template), so the test never hits its skip; it runs and finds zero patterns, which is a **genuine non-vacuous pass** because a runner truly has no local-only excludes. Confirmed `--git-common-dir` resolves to the shared `.git` (541-byte exclude) rather than the per-worktree stub. Appending an unmirrored pattern turned it red.
- **The three "restated elsewhere" claims** for the cut `Sources` bullet — all three found and quoted (lines 36, 57–59, 100–101). Dedup, not content loss.
- **Every byte figure**: `design-system-consistency` 3→139 (8189→8053), `token-naming-conventions` 657→560 (7535→7632). Exact.
- **The suite**: fresh pyenv-3.11.8 venv, `1046 passed`, exit code captured into a variable before any pipe.

## Corrections to this PR's own claims

**1. "the tightest file in the catalog" is wrong.** On `origin/dev`, `design-system-consistency` had 3 bytes — but **`test-discipline` had 1**, and it is untouched by this PR. The fix is unaffected; the superlative isn't mine to keep.

**2. The calibration tallies are not reproducible from the tree** — no calibration script is committed. The panel wrote an independent replay using this PR's own shipped `Loss` class and the module docstring's own recipe. **Control: it finds both named cases exactly** — `designing-elite-ui` (loss 36, gain 328) and the #231 lens-fill (100/95/79) — so the replay isn't broken. But it measures **`--all`: 20 pairs / 649 file-revisions** (claimed 19/170) and **`origin/dev`: 16 / 159** (claimed 7/99).

The `--all` gap sits inside the code's own "checkout-dependent" caveat. **The `origin/dev` gap does not** — that caveat claimed a 170→172 shift for two branch tips, and `origin/dev`'s tip is exactly this branch's merge-base. I cannot explain 16-vs-7 from the tree. The mechanism is independently verified correct, so this is a **reproducibility** gap, not a correctness one — but a number nobody can re-derive is the thing `AGENTS.md` says not to write.

**3. The gate's real guarantee, stated as a number:** because `gross_loss > CHURN_FLOOR` is strict, **exactly 20 masked words per scope never fire, regardless of how large the offsetting gain is.** Demonstrated with a synthetic file deleting 20 prose words and adding 999: `net 0, churn {}, exit 0`. Inherent to any floor (`FLOOR` has the identical property below its own value) and it is the disclosed tradeoff — but it deserves to be written down as a figure rather than left implicit.

**4. "Routinely bypassed" — refuted.** The 20 `--all` fires trace to ~10 distinct historical commits, roughly **8%** of the ~123 commits that ever touched a `SKILL.md`. A real tax on legitimate rewrites, as the code discloses. Not routine, so the gate is not defeated.

## Two findings filed rather than left in prose

Both were disclosed only in the PR body and commit message. **A commit message is not a queue.**

- **#264** — `skill-census`'s `KEY_MATCH`/`EXISTS` sites swallow a `gh` failure the same way, on the **per-verdict** path that runs far more often than the digest lookup. #250 is still correctly closed: its own text scopes itself to the two sites this PR fixed and never names these.
- **#265** — `spacing-system`'s masked prose loss, **confirmed live** against the shipped detector: `--base 602bb6c2ea --head d3963dd4a9` → `lost 105 words (55 code, 50 prose)`, exit 1, while the ledger row declares 55.

## Review posture

Per protocol §8.3 this panel is a gate **I** administer — tamper-evident, not independent. `clud-bug-review` is `neutral`, the repo's disclosed pre-existing outage. Nothing here is independently verified.

PR #266, opened 2026-08-24
Anchor the frontmatter closing delimiter — a line that merely starts with dashes was closing the block

thrillmot:
Found by **protocol#103's fifth-pass panel**, in code this repo had already merged (#262). Fixes the implementation, not the spec — SPEC §2.1 is correct.

## The defect

§2.1 step 2 says the frontmatter block *"closes at the first subsequent line that is **exactly** `---`."* The regex did not require exactness:

```python
FRONTMATTER_RE = re.compile(rb"^---\n(.*?)\n---", re.DOTALL)
```

Nothing forces the closing `---` to be followed by a newline or EOF, so `\n---` also matches the **leading three characters** of `----` or `--- not a close`. The block ends one character early; an identity line after that point sits in what this module treats as *body*, is never elided, and the digest moves.

**Two conforming implementations, identical bytes, different digests** — the single thing `digest` exists to prevent. Reproduced before fixing:

```
spec-faithful : 92dfcd7a1f8c
reference impl: 03a00b6b4edb
AGREE: False
```

## The fix, and why a lookahead

```python
FRONTMATTER_RE = re.compile(rb"^---\n(.*?)\n---(?=\n|\Z)", re.DOTALL)
```

A **lookahead**, not a match. Callers slice the file on `m.end()`; consuming the newline would move a byte across that seam for every conforming file. This way conforming files are byte-for-byte unaffected.

**No published digest moved** — checked against `origin/dev`'s `docs/skill-versions.json`:

```
skills whose published digest moved: 0
```

## Why fix rather than file

No real skill triggers it today — **0 of 56**, verified. But `skillforge/SKILL.md` already carries four bare `---` lines, so the shape is one plausible edit away. And `digest` is the subscriber-facing identity: the failure is silent, and a wrong answer is indistinguishable from a right one.

## Tests and mutation

Two regression tests assert **through `digest()`**, not by matching the pattern — a test on the regex passes its own mutation and still goes green when the bug ships again.

```
restore the unanchored pattern (grepped in tree at line 90)
  → 2 failed, 340 passed
restore                        → byte-identical → 342 passed
```

Gates: `validate_skills` 0 · `stamp_versions --check` 0 · `gen_skill_directory --check` 0.

## The part worth keeping

**The SPEC was verified against this implementation and still did not catch it** — both were read by the same pair of eyes, so the shared assumption was invisible. The divergence only surfaced when a panel wrote a **third** implementation from the prose alone and compared. That is the argument for writing the independent implementation rather than reviewing the existing one.


thrillmot:
## Four passes, and every one found my own remediation incomplete in the same shape

Recording the sequence, because the pattern is the finding:

| pass | what it found |
|---|---|
| 1 | The fix **made things worse** — anchored `skill_version.py`, left `skills_current.py` stale, breaking a *tested* invariant that had held while both were identically wrong |
| 2 | The sweep claimed completeness on a **symbol grep** that could not find inlined logic — **six** owners, not four |
| 3 | Guarded `_trigger_surface`'s call site, left `_split`'s. Proved it by mutating line 615: **all 39 tests stayed green** |
| — | (each fix verified, each next pass found the next gap) |

The generalisation worth keeping: **search for the behaviour, not the symbol — and control-test the search.** A grep for `FRONTMATTER_RE` cannot find `text.find("\n---", 4)`.

## The final delta, and why it is not going to a fifth panel

```
$ git diff --name-only 66490d1..HEAD | grep -v '^tests/' | grep -v '^docs/'
   (empty — tests and derived docs only)
```

One test function, 32 lines. **No production code.** Its only failure mode is being decorative — which is precisely what was verified, against the pushed state:

```
mutated line 615 (the _split site), confirmed landed in the tree
  → 1 failed, 39 passed
restored
  → 40 passed
```

A fifth pass on a mutation-proved test would be process theatre. **Stating plainly that this delta is unpanelled**, rather than letting a merge imply otherwise.

## One method note, and one correction

**Method:** my first attempt at that mutation asserted there was exactly one occurrence of the target line and **found two** — the 4-space `_split` site is a substring of the 8-space `_trigger_surface` one. The assertion is the only reason I mutated the right line instead of reporting a false green. Target by line number when two call sites differ only by indentation.

**Correction:** I reported a recurring `docs/file-structure.md` root-label leak and "fixed" it twice. It was only ever in my **working tree** — the committed and pushed branch always carried `agent-skills`, and the file is not in this PR's diff at all. I was checking the local file instead of `HEAD`: the same discipline failure as trusting a claim over the tree, pointed at myself.

The underlying generator defect is real and is agent-skills#221's open half — `logmind file-structure --write` derives the root label from the working directory's basename, so never run it from inside a differently-named worktree.

## Verification

0 published digests moved · `validate_skills` 0 · `gen_skill_directory --check` 0 · `stamp_versions --check` 0 · fast suite 1089.

PR #268, opened 2026-08-24
thrill-skills on npm: ship the catalog so Dependabot becomes the update trigger

thrillmot:
Publishes the catalog as **`thrill-skills`**, so Dependabot becomes the update trigger for any repo with a `package.json`. **Nothing is published by this PR** — the first real publish is a separate, deliberate act.

## The design: this package ships CONTENT, not tooling

The `skills` CLI is **vercel-labs'**, not ours, and it already has `experimental_sync` which bridges `node_modules` → agent directories. So there is no postinstall here, no JS reimplementation of the Python gates, and no `bin`. Writing one would put a second owner on a behaviour that already has one.

## Proved end to end, with a real tarball — not a dry run

1. `npm pack` → a real `thrill-skills-0.1.0.tgz`
2. `npm install <tarball>` in a scratch dir **outside the repo** → 56 × `node_modules/thrill-skills/skills/<slug>/SKILL.md`, **byte-identical to source** (`diff` confirmed)
3. `npx skills experimental_sync -y -a claude-code` → **"Found 56 skills in node_modules"**, all 56 written to `.agents/skills/<slug>/SKILL.md`

**Two `⚠ Skipped … ENOENT` lines appear and are not a defect.** They are the CLI's own negative probes for a root-level `SKILL.md` before it descends into `skills/<slug>/` — confirmed by reading the CLI's discovery code *and* cross-checked against a real published package with the same layout (`dotenv`). Worth naming, because a warning nobody explains is how a real one gets ignored later.

## The tarball, measured

**62 files, 187.1 kB packed / 516.0 kB unpacked** — exactly `skills/` (56 `SKILL.md` plus one `references/` subtree), `LICENSE`, `README.md`, `package.json`. **No `docs/`, `.github/` or `tests/` leaked in.** A package that accidentally ships its own test suite is a real defect, so this was measured rather than assumed.

## Two decisions, argued in the files rather than only here

**Package version is independent of per-skill semver, and monotonic.** It tracks *publishes of the tarball*, not any one skill's content — that is what `version:` / `digest:` / `origin:` frontmatter already owns per skill. Advanced by hand (bump, tag `vX.Y.Z`), same as `clud-bug`'s own package. **Nothing auto-tags a `skills/` merge yet** — flagged as an implication rather than built, since it is outside this brief.

**`skill-versions.json` is deliberately NOT shipped.** That index answers *"is my copy, obtained some other way, stale?"* — which does not apply to a copy that arrived by `npm install` plus a Dependabot bump. Its own `_readme` also describes `main`-branch CI recomputation that means nothing divorced from the git repo. Reversible if we find a consumer who needs it.

## Publishing

OIDC **Trusted Publishing** (`id-token: write`, `--provenance`) — **no `NPM_TOKEN` secret**, matching the pattern already proven in `thrillmade/clud-bug`. Triggers only on a `vX.Y.Z` tag or explicit `workflow_dispatch`; **`pull_request` is not in the `on:` block at all**, so a PR cannot publish. Re-runs `validate_skills.py` at the tagged ref as cheap insurance. `actionlint` clean.

## Caught before committing, worth recording

The lane ran `logmind file-structure --write`, which regenerated `docs/file-structure.md` — a derived, `main`-only doc whose branch edit trips a blocking gate. It caught this itself and reverted with `git checkout origin/dev -- docs/file-structure.md` **before** committing, so the final diff never touches it.

That is the same trap that has bitten three lanes today. Confirmed clean here: derived docs in this diff = **0**.

## Verification

`validate_skills.py` 56 · `stamp_versions.py --check` 56 · `gen_skill_directory.py --check` current · `logmind check-links` no broken or orphan links · `actionlint` clean · **full `pytest tests/ -q` → 1218 passed**, including both mutation files.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -42,6 +42,32 @@ jobs:
       contents: read
       id-token: write    # OIDC for npm Trusted Publishing + provenance
     steps:
+      # Guard the workflow_dispatch path. `inputs.tag || github.ref` falls
+      # back to github.ref, which on workflow_dispatch is the BRANCH the
+      # dispatch was fired from -- typically main, not a tag. A stray
+      # dispatch with an empty tag would therefore publish main's HEAD as a
+      # release. `required: true` covers the web form; it is not a runtime
+      # guarantee for an API- or CLI-triggered dispatch, so this checks at
+      # run time. Ported from thrillmade/clud-bug's own npm-publish.yml,
+      # which carries this guard for exactly this reason -- an earlier
+      # version of THIS file cited that workflow as its pattern and then
+      # dropped the guard the pattern exists for.
+      # Tag-push triggers skip it (event_name == 'push').
+      - name: Validate dispatch inputs
+        if: github.event_name == 'workflow_dispatch'
+        env:
+          TAG_INPUT: ${{ inputs.tag }}
+        run: |
+          set -euo pipefail
+          if [[ -z "${TAG_INPUT}" ]]; then
+            echo "::error::workflow_dispatch requires an explicit tag. Without one this would publish the branch HEAD, not a release."
+            exit 1
+          fi
+          if [[ ! "${TAG_INPUT}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "::error::tag '${TAG_INPUT}' is not vX.Y.Z shaped."
+            exit 1
+          fi
+
       - name: Checkout the tag
         uses: actions/checkout@v7
         with:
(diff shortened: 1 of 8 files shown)


thrillmot:
## Panel: MERGE, 0 blocking — and it reproduced the whole chain itself

It didn't take the end-to-end claim on trust: it ran its own `npm pack`, its own install, its own `experimental_sync`, and got the identical 56 + the identical two `ENOENT` lines. Then it cross-checked those against `dotenv@17.4.2` and confirmed a real published package shows the same shape.

It also went further than I did on the `skill-versions.json` exclusion: it **recomputed the digest algorithm by hand** on `skills/skillforge/SKILL.md` and got `ea1de66ab57a`, matching that file's own `digest:` frontmatter. So a consumer *can* detect a local edit using only what ships. The argument holds because it was tested, not because it sounded right.

## Both non-blocking items are fixed here, not deferred

The panel said these don't block merging to `dev` since this PR publishes nothing — true — and should be fixed **before the first real publish**. That is a promise to remember, and remembering is the thing that fails. Fixed now.

### 1. `workflow_dispatch` had no runtime guard — and this is the sharp one

`inputs.tag || github.ref` falls back to `github.ref`, which on a dispatch is **the branch it was fired from**, typically `main`. A stray dispatch with an empty tag would publish main's HEAD as a release.

**This file cited `thrillmade/clud-bug`'s workflow as its pattern and then dropped the guard that pattern exists for.** clud-bug's own comment says it outright:

> *"Without this check, a stray 'publish' dispatch with no tag would resolve `inputs.tag || github.ref` to github.ref — which on workflow_dispatch is the BRANCH the dispatch was fired from… The result would be a silent main-branch publish instead of a release publish."*

`required: true` covers the web form. It is not a runtime guarantee for an API- or CLI-triggered dispatch.

### 2. Nothing compared the tag to `package.json`

npm refuses to re-publish a version, so a `v9.9.9` tag on a `0.1.0` package publishes **`0.1.0`** — and the git tag and the registry disagree permanently, with nothing to catch it.

### Both tested in isolation, not reasoned about

```
empty tag              -> REJECT     v0.1.0                 -> accept
"main"                 -> REJECT     v9.9.9 on pkg 0.1.0    -> REJECT
                                     v0.1.0 on pkg 0.1.0    -> accept
```

`actionlint` clean, YAML parses.

## One thing the panel found that is worth keeping in view

`files:` **bypasses `.gitignore`** — a gitignored file physically inside `skills/` would be packed. It verified this empirically in a scratch replica. It is closed *in practice* only because the sole publish path is `actions/checkout@v7` against a tagged ref in fresh CI, where there is no working-tree state to leak. Nothing in `package.json` prevents it.

Currently clean (`git status --ignored=matching skills/`), and worth remembering the day anyone considers a manual `npm publish` from a laptop.

## Also verified by the panel, independently

Zero credential-shaped or client-specific content in `skills/` — grepped for AWS/GitHub/OpenAI key shapes, private-key headers, and confidential markers, **control-tested against known-present strings** so the zeros are real. `NPM_TOKEN` appears only in comments; no `secrets.*` anywhere. `permissions: {}` at workflow level with `id-token: write` scoped to the one job. Tarball re-measured: 62 files, 187,131 bytes — exact match.

